PR #79, opened 2024-02-03
Fixed: the routing from the query page to the route list page

### LINKED ISSUE #62: Routing from Run List page to Run Details page
## Current behavior

When routing from the Run list page (home) to the detail page the user is pushing a page onto the stack. When going back from the run details to the list page, it feels like the user is adding another page to the stack when it should actually be popping from the stack. I think we need to verify once that the back action is mapped correctly.

I am judging this based on the page transitions FYI

ymaheshwari1:
### Related Issues

Closes #62 

After: 

image

(after font-weight update)

### DIFF
--- a/src/views/BrokeringQuery.vue
+++ b/src/views/BrokeringQuery.vue
@@ -5,10 +5,7 @@
         <div class="menu">
           <ion-item lines="none">
             <ion-label>{{ currentRouting.routingName }}</ion-label>
-            <ion-chip slot="end" outline @click="router.push(`/tabs/brokering/${currentRouting.routingGroupId}/routes`)">
-              {{ getRouteIndex() }}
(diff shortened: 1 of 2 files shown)


dt2patel:
Are we required to use ion back button if we want to manage routing correctly? I feel like we should be allowed to add a corresponding pop/back action to any element we want right?

ymaheshwari1:
> Are we required to use ion back button if we want to manage routing correctly? I feel like we should be allowed to add a corresponding pop/back action to any element we want right?

We have method to go to specific page using `router.go(number)` in the route history, but as we have support to move to the settings page from the query page, if the user moves to the settings page a new route entry gets added in the history and when clicking back from query page, the user gets redirected to the settings page, as user is coming from the settings page.

PR #176, opened 2024-04-10
Fixed: operator when selecting proximity filter

ymaheshwari1:
### Related Issues

When applying proximity filter the operator passed is `equals` but we don't need the distance calculation logic to check for exact distance.

Changed the operator for proximity filter to `less-equals` as we need to check for all the facilities withing that specified distance.


Adameddy26:
Verified and working fine in release v1.1.0.

PR #202, opened 2024-04-22
Fixed: case to have a default value for cronExpression when updating group status and cron expresssion is not present

ymaheshwari1:

When changing the status of a group to active, and if the schedule is not set for the group then making the default value for the schedule to EveryDay at mid-night


Adameddy26:
Verified and working fine in release v1.2.0. 
https://jam.dev/c/58e9a582-ecf0-4dfb-a383-2bb1ce18e263

PR #214, opened 2024-05-13
Fixed: issue of duplicate filters getting applied on the rules

ymaheshwari1:
### Related Issues

This issue occurs when we apply a new condition in the rule and update the same condition without refresh.

Fetched the latest information after updating a rule so to pass the seqId when updating the same filter again


Adameddy26:
Verified and working fine in release v1.3.0.

Issue #248, opened 2024-07-31
Exclude filter options


dt2patel:
## What is the motivation for adding/enhancing this feature?

Add option to exclude in routing filters for both orders and inventory rules:

<img width="480" alt="Screenshot 2024-07-31 at 10 28 29 AM" src="https://github.com/user-attachments/assets/dc88b98b-744b-4910-b1ee-ad623411d3b9">
<img width="517" alt="Screenshot 2024-07-31 at 10 28 44 AM" src="https://github.com/user-attachments/assets/e75d58ed-416b-4471-8f2a-d1e0f3005899">

If a filter is added as an exclusion operator, then it should have an "excluded" secondary line in "danger" color

Adameddy26:
Verified the changes and are working fine in release v1.7.0. Hence, closing this issue.

PR #262, opened 2024-09-30
Implemented: support to exclude order and inventory filters when configure a route(#248)

ymaheshwari1:
### Related Issues

#248 

After:

image

image


ymaheshwari1:
For order filters, we have added all the current filters in the excluded list, and we have not handled the case to not allow the user to select those options that are already selected in its corresponding exclude/include filter. Here we have considered that user configuring the rules will be aware of the conflicts that may occur when selecting same filter as exclude and include both with similar values.

For inventory filter, we have added exclude filter option only for the facilityGroup filter, as all the other filters does not need to have the exclude check. For the facility group filter, we have handled the case to not allow the select the same value in the exclude and include filter.

Issue #169, opened 2024-04-08
Inaccurate Cursor Pointer Display on Selected Filter in Inventory Rule


Adameddy26:
## Expected behavior
The cursor should either stay default or signify that the filter card is not clickable, as there's no action associated with clicking it.

## Steps to reproduce the issue
Login the application using valid credentials.
Open any run and then get inside routing to see the rules.
Select a filter to apply to the inventory rule.
Hover the cursor over the selected filter card to observe the issue.

## Environment details
App version: MAIN-E3C8C75
App environment: Dev

## Additional information
https://jam.dev/c/a70cb599-79ae-485c-b379-e4c7068a2de6

dt2patel:
I guess what we're trying to say here is that items in the filter card for inventory are clickable because of the click action attached to them. In my experience when adding the proximity option, clicking the item some times results in the select being triggered and other times nothing happens.

PR #295, opened 2025-04-14
Fixed: Prevented pointer cursor and click events on ion-item (#169)

R-Sourabh:
### Related Issues

#169 

- The ion-item contains an ion-select element within an ion-chip. The presence of this `interactive control` (ion-select) is the reason why Ionic dynamically adds the `.item-has-interactive-control` class to the ion-item, resulting in the pointer cursor. 
- Added `cursor: default` and `pointer-events: none` on the `ion-item` to prevent the pointer cursor and click events on its background, while `pointer-events: auto` was applied to the nested `ion-chip` to maintain its interactivity.


ymaheshwari1 (approved):

ymaheshwari1:
@R-Sourabh Cross check on any other place where this needs to be handled

dt2patel:
This seems like a manual solution to a problem that has impact in many places. It's not a high priority issue so I'd like us to take time and think about how ionic has implemented this, is there something wrong with that default behavior? 

If so, why did they choose to implement it that way, why doesn't it work for us? Maybe we're trying to use it wrong. Could be my fault for not understanding the component and creating bad design.

Issue #36, opened 2024-01-23
Implement support to let user define the custom scheduling freqency


ymaheshwari1:
## What is the motivation for adding/enhancing this feature?

In some cases, user may need to define some custom frequency, and for this case we need to have some kind of support to let user easily define the frequency.

## What are the acceptance criteria?

- [ ] ...

## Can you complete this feature request by yourself?

- [x] YES
- [ ] NO

## Additional information

Use Cron Expression Generator for defining custom frequency. Below is the list of possible options:

https://medium.com/js-dojo/getting-start-with-vue-scheduler-b9a91bd4d14e
https://morioh.com/a/eea3014c18e0/easy-cron-editor-for-vuejs
https://karoletrych.github.io/vue-cron-editor/
https://github.com/abichinger/vue-js-cron

ymaheshwari1:
For converting cron-expression to human readable expression, we can use this npm package: https://www.npmjs.com/package/cronstrue

dt2patel:
Moving to dev complete so that @Adameddy26 can pick for testing

dt2patel:
There is an issue that day of week cron expressions do not get accepted by Moqui, life every 2 hours from Monday to Friday cannot be submitted from app. @shubham-namdeo please test this case

shubham-namdeo:
> There is an issue that day of week cron expressions do not get accepted by Moqui, life every 2 hours from Monday to Friday cannot be submitted from app. [@shubham-namdeo](https://github.com/shubham-namdeo) please test this case

sir, found that this issue exists because of cron expression parsing error. Hence, reopening the issue.
https://jam.dev/c/ad19a059-d93a-479b-830e-95c24edea139

ymaheshwari1:
Here the sample expression used is defining both day-of-week and day-of-month which is an invalid cron expression, thus a valid expression will be: 0 15 */2 ? * 1-5

dt2patel:
Add info link to library syntax of CRON expression

Nirvan-Bobde30:
@dt2patel do we need to add the link of library on UI?

dt2patel:
Yes. Add it as an info icon to the end of the input field

PR #308, opened 2025-08-07
Implemented : support to let user define the custom scheduling frequency (#36)

Nirvan-Bobde30:
### Related Issues

#36 

- Added a info icon to the end of the input field .

image

Updated UI after changes
image

### DIFF
--- a/src/components/ScheduleModal.vue
+++ b/src/components/ScheduleModal.vue
@@ -14,7 +14,7 @@
     <ion-list>
       <ion-item>
         <ion-input label-placement="floating" :label="translate('Expression')" v-model="expression"></ion-input>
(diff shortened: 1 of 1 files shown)


src/components/ScheduleModal.vue
```diff
@@ -138,4 +141,7 @@ async function saveChanges() {
     });
   return alert.present();
 }
+function goToCronHelp() {
+  window.open('https://www.freeformatter.com/cron-expression-generator-quartz.html',"_blank")
```
dt2patel: instead of functions, why not use button href property?
https://ionicframework.com/docs/api/button#href

dt2patel (changes requested):

PR #352, opened 2026-06-09
chore(toolchain): pinia-3 / vue-tsc 3.x migration + type fixes

toaditi:
## What

Ports the **Pinia-3 / Vue-3.5 toolchain migration + type fixes** (developed while integrating ATP-sourcing into the simulation/outcomes branch) onto `feat/merge-atp-sourcing-into-routing`.

## Why

The ATP branch pins `typescript: catalog:` → **TS 6.0.2**, which silently breaks Pinia-3 `defineStore` getter/action inference under `vue-tsc` (typecheck effectively no-ops). This PR adopts the known-good working combo so the app actually type-checks.

## Changes (24 files, +37/−34)

- **`package.json`** — pin `typescript@5.9.3`, add `vue-tsc@3.3.4`.
- **`tsconfig.json`** — `moduleResolution: bundler`, drop `baseUrl`, `paths` → `./src/*`.
- **`src/store/userStore.ts`** — add additive getters `getCurrentTimeZone` + `getAvailableTimeZones`.
- **~20 `.vue` files** — vue-tsc 3.x type fixes: remove `defineProps` macro imports, `String()`/`Number()`/`?? 0` casts for never/index typing, `(): any` annotations on computeds/getters, `ref<any[]>`, `:debounce` binding.

## Scope / deliberately excluded

- ❌ All simulation / circuit work (services, UI, animation, double-login auth).
- ❌ `orderRoutingStore.fetchInventoryRuleInformation` getter-style refactor — it **drops the API fetch + `filterSortDesc` handling** and depends on a `getCurrentRule` getter **not present on this branch**, so it's a behavioral change, not a safe migration fix. Left out intentionally.

## Verification

`npx vue-tsc --noEmit -p tsconfig.json` → **0 errors in app source**. The only remaining error is a pre-existing `defineProps` import conflict in the shared `common/components/ImageModal.vue` (outside this repo, untouched here).

### DIFF
--- a/src/store/userStore.ts
+++ b/src/store/userStore.ts
@@ -197,7 +197,8 @@ export const useUserStore = defineStore('user', {
       }
     },
     async getAvailableTimeZones() {
-      return this.fetchAvailableTimeZones()
+      await this.fetchAvailableTimeZones()
+      return this.timeZones
     },
     async fetchAvailableTimeZones() {
       if (this.timeZones.length) return;
--- a/src/store/userStore.ts
+++ b/src/store/userStore.ts
@@ -196,10 +196,6 @@ export const useUserStore = defineStore('user', {
         commonUtil.showToast(translate("Time zone updated successfully"));
       }
     },
-    async getAvailableTimeZones() {
-      await this.fetchAvailableTimeZones()
-      return this.timeZones
-    },
     async fetchAvailableTimeZones() {
       if (this.timeZones.length) return;
       try {
--- a/src/store/userStore.ts
+++ b/src/store/userStore.ts
@@ -39,6 +39,9 @@ export const useUserStore = defineStore('user', {
     getPwaState(state) {
       return state.pwaState;
     },
+    getCurrentTimeZone(state): any {
+      return state.current?.timeZone
+    },
     hasPermission: (state: any) => (permissionId: string): boolean => {
       const permissions = state.permissions;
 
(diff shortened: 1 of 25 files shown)


toaditi:
Addressed the gemini-code-assist review comments:

1. **`getCurrentTimeZone` → `string | undefined`** (accepted) — this surfaced a real consumer break: `DxpTimeZoneSwitcher.vue` passes it to `commonUtil.getCurrentTime(zone: string)`. Added a `?? ''` fallback at the call site so it type-checks while preserving behaviour (`setZone('')` resolves to the default zone, same as the previous `undefined` path).
2. **`getAvailableTimeZones`** — now `await this.fetchAvailableTimeZones()` then `return this.timeZones`, so it actually returns the list.
3. **`routingsForReorder`** — typed `ref<Route[]>` instead of `ref<any[]>`.

Verified: `npx vue-tsc --noEmit -p tsconfig.json` → **0 errors in app source** (only remaining error is a pre-existing `defineProps` import conflict in the shared `common/components/ImageModal.vue`, untouched by this PR).

src/store/userStore.ts
```diff
@@ -193,6 +196,10 @@ export const useUserStore = defineStore('user', {
         commonUtil.showToast(translate("Time zone updated successfully"));
       }
     },
+    async getAvailableTimeZones() {
```
ymaheshwari1: Instead of having a wrapper here, we can directly call fetchAvailableTimeZones from DxpTimeZoneSwitcher

ymaheshwari1 (changes requested):

src/store/userStore.ts
toaditi: Done in 08900cf — removed the `getAvailableTimeZones` wrapper and `DxpTimeZoneSwitcher` now calls `userStore.fetchAvailableTimeZones()` directly. vue-tsc clean.

PR #354, opened 2026-06-12
feat(sim): brokering simulation — outcome metrics, what-if variations, per-routing drill-down

toaditi:
## Summary
- **Simulate tab** (behind `VITE_SIMULATION_ENABLED`): run brokering simulations against the sim backend, with outcome metric panels (fulfillment mix, stockouts, expedited, composite score, tradeoff chart) and a past-simulations history list
- **What-if variations**: H2 variation editor on the canvas flow with persist-on-save via `PUT /config`, run controls, and a parent-vs-variation compare table with per-routing result join
- **Per-routing drill-down**: clickable result rows open a run-result detail modal; `traceRollup` turns rule attempts into a plain-English narrative, plus outcome counts and facility rollups from order traces
- **Two-instance dev support**: the sim Moqui has its own `api_key` auth (`SimAuthService`) and the Simulate editor reads reference data from a dedicated `simReferenceStore`, so the OMS and sim backends never share request paths or in-memory state (see `docs/simulation-login-technical.md`)
- **RoutingGroupService extraction**: one shared fetch+normalize implementation for the OMS and simulate paths
- **Circuit**: Settings "Show Circuit" toggle; the branch lineage also carries the Circuit assistant / route-assistant work it builds on

## Config
- `VITE_SIM_API_BASE_URL` — sim backend base URL (UAT); when blank the app stays single-instance
- `VITE_SIM_URL` / `VITE_SIM_USERNAME` / `VITE_SIM_PASSWORD` — local two-instance dev only
- The simulation backend is moving from `order-routing` to the new `ai-routing` component; the REST contract is stable and the `order-routing/` URL prefix is the single swap point

## Test plan
- [x] `npm run build` green
- [x] 23 plain-node test files via `npx tsx` — all pass
- [x] 15 vitest files via `npx vitest run` — 46/46 assertions pass
- [ ] Manual: run a group simulation on UAT, open the per-routing drill-down, save + run an H2 variation

## Notes for reviewers
- Branch merges cleanly into `main` (verified with `git merge-tree` — no conflicts). The lineage includes the pinia-3 toolchain commits that were squash-merged in #352; the duplicated content auto-resolves.


src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const ENDPOINT = "/brokering-runs-list-inquiry";
+
+export async function requestBrokeringRunsListInquiry(
+  prompt: string,
+  manifest: PageCapabilityManifest,
+  conversationHistory: DraftConversationMessage[] = []
+): Promise<BrokeringRunsListInquiryResult> {
+  const mastraUrl = (import.meta.env.VITE_MASTRA_URL || "http://localhost:4111").replace(/\/$/, "");
```
ravilodhi: Write a utility function getMastraUrl and use it across the files.

src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,57 @@
...
+  // The order-routing/facility-changes endpoint lives on Moqui (Maarg), not OFBiz/OMS,
+  // so the assistant tool needs the Maarg base URL.
+  const omsBaseUrl = commonUtil.getMaargURL() || commonUtil.getOmsURL();
+  const authToken = cookieHelper().get("token");
```
Banibrata-Manna: Please use commonUtil's getToken here

src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,57 @@
...
+  let response: Response;
+  try {
+    response = await fetch(`${mastraUrl}${ENDPOINT}`, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        prompt,
+        conversationHistory,
+        pageCapabilityManifest: manifest,
+        ...(omsBaseUrl ? { omsBaseUrl } : {}),
```
ravilodhi, line 31: Why do we need to pass omsBaseUrl and token to mastra?

src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,57 @@
...
+  conversationHistory: DraftConversationMessage[] = []
+): Promise<BrokeringRunsListInquiryResult> {
+  const mastraUrl = (import.meta.env.VITE_MASTRA_URL || "http://localhost:4111").replace(/\/$/, "");
+  // The order-routing/facility-changes endpoint lives on Moqui (Maarg), not OFBiz/OMS,
+  // so the assistant tool needs the Maarg base URL.
+  const omsBaseUrl = commonUtil.getMaargURL() || commonUtil.getOmsURL();
+  const authToken = cookieHelper().get("token");
+  let response: Response;
+  try {
+    response = await fetch(`${mastraUrl}${ENDPOINT}`, {
```
ravilodhi: Instead of using fetch, use the 'client' method from @common (accxui remoteApi)

src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,57 @@
+import { commonUtil, cookieHelper } from "@common";
+import type { DraftConversationMessage, PageCapabilityManifest } from "@/services/DraftAssistantService";
+
+export type BrokeringRunsListInquiryResult = {
+  message: string;
+  questions: string[];
+  summary: string;
+};
+
+const ENDPOINT = "/brokering-runs-list-inquiry";
```
ravilodhi: Use the client function from @common (accxui), which supports baseURL and url. Set baseURL to the Mastra URL and url to the specific endpoint path.

src/services/CircuitDraftFeedbackService.ts
```diff
@@ -0,0 +1,37 @@
+export type DraftFeedbackType = "approved" | "revision_requested" | "rejected";
```
ravilodhi: his file seems to contain only utility/helper functions. The service layer should be responsible for calls to external systems. Payload preparation and response transformation logic are fine here, but functions that do not interact with external systems should be moved to a utility/helper file.

src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,57 @@
+import { commonUtil, cookieHelper } from "@common";
+import type { DraftConversationMessage, PageCapabilityManifest } from "@/services/DraftAssistantService";
+
+export type BrokeringRunsListInquiryResult = {
```
ravilodhi: Create a separate file under src/types/ and move all the type definitions there. Import and use them from that file wherever needed.

src/store/simulationStore.ts
```diff
@@ -0,0 +1,348 @@
...
+        const ids = j.variantLabels.map((label, n) => {
+          const id = `${j.jobId}:${n}`;
+          this.runStates.push({ variationId: id, label, phase: "running" as const });
+          return id;
+        });
+        return { batchIndex: j.batchIndex, ids, jobId: j.jobId };
+      });
+
+      const batchResults = await Promise.all(toRun.map((x) => this.runBatch(x)));
+      try { this.results = mergeVariationResults(batchResults); } finally { this.isRunning = false; }
```
Banibrata-Manna: We should wrap mapping of jobs too in a try block here.
Any error before this can leave the isRunning flag as true. If this is used anywhere to show the loading state, then the UI will get stuck in the loading state.

src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,57 @@
...
+  message: string;
+  questions: string[];
+  summary: string;
+};
+
+const ENDPOINT = "/brokering-runs-list-inquiry";
+
+export async function requestBrokeringRunsListInquiry(
```
ravilodhi: Also, instead of exporting each service function individually, export them as part of a service object, like this:

export const BrokeringRunAssistantService = {
  requestBrokeringRunsListInquiry,
};

Then use BrokeringRunAssistantService.requestBrokeringRunsListInquiry wherever it is called. This way, you only need to import the service object, and you can access its methods using dot notation.

src/store/simulationStore.ts
```diff
@@ -0,0 +1,348 @@
...
+        this.batchProgress = batches.map((_, i) => zeroedBatch(i));
+
+        // Submit every batch first (instant responses) so we have all jobIds to persist up-front.
+        SimulationJobStore.clearJobs(this.routingGroupId);
+        const submitted = await Promise.all(batches.map(async (variants, i) => {
+          const ids = idBatches[i];
+          this.setVariationPhase(ids, "submitted");
+          try {
+            const jobId = await submitBatch({ routingGroupId: this.routingGroupId, variants });
+            return { batchIndex: i, ids, jobId, variantLabels: variants.map((v) => v.label) };
```
Banibrata-Manna, line 292: isRunning is never reset on success or any errors in inner try-catch(s)

src/services/BrokeringRunsAssistantService.ts
ravilodhi: Example: await client({
      url: "admin/user/profile",
      method: "get",
      baseURL: getMastraUrl,
      headers: {
        'Content-Type': 'application/json'
      },
      data: {
       }
})

src/components/simulation/SimulationCanvas.vue
```diff
@@ -0,0 +1,1735 @@
...
+  const alert = await alertController.create({
+    header: translate("Save as new variation"),
+    inputs: [{ name: "label", placeholder: translate("Variation name") }],
+    buttons: [
+      { text: translate("Cancel"), role: "cancel" },
+      {
+        text: translate("Save"),
+        handler: (data) => {
+          flushWorking();
+          sim.saveAsVariation(data?.label);
```
Banibrata-Manna: Please use await here.

src/services/SimulationService.ts
```diff
@@ -0,0 +1,237 @@
...
+      return { done: true, error: resp.error || "Simulation failed." };
+    case "not_found":
+      return { done: true, error: "Simulation job expired before it completed. Please re-run." };
+    default:
+      return { done: true, error: `Unknown job status: ${(resp as any).status}` };
+  }
+}
+
+// Live progress feed streams per-order events as deltas, so a tight 2.5s cadence stays cheap.
+const POLL_INTERVAL_MS = 2_500;
```
ravilodhi: Define these intervals in env file.

src/services/SimulationService.ts
```diff
@@ -0,0 +1,237 @@
...
+export interface SubmitBatchArgs {
+  routingGroupId: string;
+  variants: SimVariant[];
+  sampleCap?: number;
+}
+
+/** POST one batch (≤5 variants). Returns the jobId. Throws on non-2xx. */
+export async function submitBatch({ routingGroupId, variants, sampleCap }: SubmitBatchArgs): Promise<string> {
+  const { commonUtil } = await import("@common");
+  const resp: any = await simApi({
```
ravilodhi: Use the client function here and remove the simApi function completely.

src/services/SimulationService.ts
```diff
@@ -0,0 +1,237 @@
...
+ *  TODO: replace with a store selector populated from the sim instance's product stores. */
+export function simProductStoreId(env: Record<string, any> = import.meta.env): string {
+  return ((env && env.VITE_SIM_PRODUCT_STORE_ID) || "").trim();
+}
+
+/** REST root of the dedicated simulation Moqui (e.g. http://localhost:8075/rest/s1/), or "" when
+ *  unset. When set, reference-data and routing-group calls go to that host; when blank, they fall
+ *  through to api()'s default OMS base. Auth is always the shared OMS Bearer token. Env is
+ *  injectable for headless testing. */
+export function simBaseURL(env: Record<string, any> = import.meta.env): string {
```
ravilodhi, line 24: Create a utility function to get the simulation moqui base URL e.g. getSimMaargUrl.

src/services/SimulationService.ts
```diff
@@ -0,0 +1,237 @@
...
+export function simApiName(env: Record<string, any> = import.meta.env): string {
+  return ((env && env.VITE_SIM_API_NAME) || "order-routing").trim();
+}
+
+/** Product store id to scope the simulation's routing-group + reference-data queries to. The sim
+ *  instance's real stores (e.g. SM_STORE) differ from the OMS demo placeholder ("STORE") that
+ *  currentEComStore resolves to, so set VITE_SIM_PRODUCT_STORE_ID to a real one. Blank => caller falls
+ *  back to the OMS currentEComStore. Env is injectable for headless testing.
+ *  TODO: replace with a store selector populated from the sim instance's product stores. */
+export function simProductStoreId(env: Record<string, any> = import.meta.env): string {
```
ravilodhi, line 16: Can we use the selected product store in the routing app for this? If not then create a utility function for this.

src/services/SimulationService.ts
```diff
@@ -0,0 +1,237 @@
...
+  return ((env && env.VITE_SIM_URL) || "").trim();
+}
+
+/** Base URL for the brokering simulation / Sim-routing API. These endpoints live on a dedicated
+ *  instance (UAT: asb-sim-uat.hotwax.io), separate from the OMS/Maarg the rest of the app talks to.
+ *  This is the ONE place the host + component prefix are configured, so promoting UAT -> prod is a
+ *  config change, not a code change. The value INCLUDES the `/rest/s1/order-routing` prefix; per-call
+ *  paths are relative to it. Auth is unchanged — api() attaches the same credentials as other
+ *  order-routing admin calls. Env is injectable for headless testing (import.meta.env is the default). */
+export function simApiBaseUrl(env: Record<string, any> = import.meta.env): string {
```
ravilodhi, line 34: Same for this.

src/util/featureFlags.ts
```diff
@@ -0,0 +1,17 @@
+// src/util/featureFlags.ts
```
ravilodhi, line 1: We can minimize the number of utility files by grouping related functionality logically.

src/theme/circuit.css
```diff
@@ -0,0 +1,5 @@
+/* Global styles for Circuit module */
```
ravilodhi: Where is this styling used? I don't see any usage of it. We typically add global CSS to /theme/variables.css and keep component-specific styles scoped within the corresponding .vue file.

src/store/simReferenceStore.ts
```diff
@@ -0,0 +1,92 @@
...
+      const prefix = simApiName()
+      // Mirrors the old productStore guard: without a store there is nothing meaningful to scope the
+      // store-level slices to, and interpolating a blank id would request /productStores/undefined/...
+      if (!productStoreId) {
+        logger.warn("Skipping store-scoped sim reference fetches because productStoreId is missing.")
+      }
+      const [facilities, shippingMethods, facilityGroups, salesChannels] = await Promise.all([
+        this.fetchSimMap(`${prefix}/facilities`, { pageSize: 500 }, "facilityId"),
+        productStoreId
+          ? this.fetchSimMap(`${prefix}/productStores/${productStoreId}/shippingMethods`, { productStoreId, pageSize: 200 }, "shipmentMethodTypeId")
```
ravilodhi: Remove this fetchSimMap function, use the client function here.

src/store/simReferenceStore.ts
```diff
@@ -0,0 +1,92 @@
...
+      const failed = [facilities, shippingMethods, facilityGroups, salesChannels].some((r) => r === null)
+      this.productStoreId = failed ? "" : productStoreId
+    },
+
+    /** One reference slice from the sim instance, reduced to a map keyed by `keyField`.
+     *  Returns {} for an empty/garbage body and null when the request errored (so callers can tell
+     *  "nothing there" apart from "fetch failed" for caching decisions). */
+    async fetchSimMap(url: string, params: Record<string, any>, keyField: string): Promise<Record<string, any> | null> {
+      try {
+        const resp = await simApi({ url, method: "GET", baseURL: simMoquiUrl(), params })
```
ravilodhi: Use client here and remove simApi

src/services/CircuitLLMService.ts
```diff
@@ -0,0 +1,118 @@
...
+
+// Select a model that balances speed and quality. 
+// Hermes-3-Llama-3.1-8B is allow-listed for function calling in WebLLM.
+export const SELECTED_MODEL = "Llama-3.2-1B-Instruct-q4f16_1-MLC";
+
+let engine: MLCEngine | null = null;
+
+const CircuitLLMService = {
+  async isWebGPUSupported(): Promise<{ supported: boolean; error?: string }> {
+    if (!navigator.gpu) {
```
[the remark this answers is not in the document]
toaditi: Fixed. Added `typeof navigator === "undefined"` guard before the `navigator.gpu` check.

src/services/CircuitLLMService.ts
```diff
@@ -0,0 +1,118 @@
...
+      engine = null;
+    }
+  },
+
+  async getGPUInfo() {
+    let vendor = "Unknown";
+    let bytes = 0;
+
+    if (navigator.gpu) {
+      const adapter = await navigator.gpu.requestAdapter();
```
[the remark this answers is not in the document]
toaditi: Fixed. Added `typeof navigator !== "undefined"` guard.

src/components/simulation/SimulationCanvas.vue
```diff
@@ -0,0 +1,1735 @@
...
+} from 'ionicons/icons';
+import { computed, onMounted, onUnmounted, ref, watch } from 'vue';
+import { simulationStore } from "@/store/simulationStore";
+import { productStore } from '@/store/productStore';
+import { useUtilStore } from '@/store/utilStore';
+import { translate, logger, commonUtil } from '@common';
+import AddOrderRouteFilterOptions from "@/components/AddOrderRouteFilterOptions.vue"
+import { Rule } from "@/types";
+import AddInventoryFilterOptionsModal from "@/components/AddInventoryFilterOptionsModal.vue";
+import ArchivedRoutingModal from "@/components/ArchivedRoutingModal.vue"
```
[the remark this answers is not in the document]
toaditi: Not applicable in current code — the global Save FAB was already removed before this review. `SimulationCanvas` has no `save()` that writes to `orderRoutingStore`; all persistence goes through `saveAsVariation` / `updateVariation` only.

src/components/simulation/SimulationCanvas.vue
```diff
@@ -0,0 +1,1735 @@
...
+  color: var(--ion-color-primary);
+}
+
+.action-prompt p {
+  margin: 0;
+  font-size: 14px;
+  font-weight: 500;
+  color: var(--ion-color-dark);
+}
+
```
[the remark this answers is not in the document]
toaditi: Same as above — the `save()` function referencing `orderRoutingStore` was already absent from the current version of the file before this review.

src/draftTargets/BrokeringRulesDraftTargets.ts
```diff
@@ -0,0 +1,1202 @@
...
+
+export function buildBrokeringRulesBindings(draft: BrokeringRulesDraftRefs): DraftTargetBindings {
+  const targets = [
+    "route.statusId",
+    "selectedRule.statusId",
+    ...routeFilterDefinitions.map((definition) => `route.orderFilters.${definition.key}`),
+    ...routeSortDefinitions.map((definition) => `route.orderSorts.${definition.key}`),
+    ...inventoryFilterDefinitions.map((definition) => `selectedRule.inventoryFilters.${definition.key}`),
+    ...inventorySortDefinitions.map((definition) => `selectedRule.inventorySorts.${definition.key}`),
+    "selectedRule.partialAllocation",
```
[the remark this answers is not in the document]
toaditi: Not applied — the store always initializes `inventoryFilters` as `{ ENTCT_FILTER: {}, ENTCT_SORT_BY: {} }` (lines 572-574), never as a flat array. Accessing `.ENTCT_FILTER` on a flat array could theoretically return `undefined` if someone passed raw backend data directly, but that path doesn't exist in the current codebase. Added a note here; happy to add the normalization if you'd prefer defensive handling.

src/store/orderRoutingStore.ts
```diff
@@ -98,33 +111,157 @@ export const orderRoutingStore = defineStore('orderRouting', {
...
+              delete filter.routingRuleId;
+            });
+            rule.actions?.forEach((action: any) => {
+              delete action.routingRuleId;
+            });
+          });
+        });
+      } else {
+        transformedPayload.routings.forEach((routing: any) => {
+          if(validate(routing.orderRoutingId)) delete routing.orderRoutingId
```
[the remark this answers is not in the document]
toaditi: Fixed. Added truthy guard: `if(routing.orderRoutingId && validate(routing.orderRoutingId))`.

src/store/orderRoutingStore.ts
```diff
@@ -98,33 +111,157 @@ export const orderRoutingStore = defineStore('orderRouting', {
...
+          routing.rules?.forEach((rule: any) => {
+            if(validate(rule.routingRuleId)) delete rule.routingRuleId
```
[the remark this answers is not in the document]
toaditi: Fixed. Added truthy guard: `if(rule.routingRuleId && validate(rule.routingRuleId))`.

src/services/SimulationHistoryCache.ts
```diff
@@ -0,0 +1,98 @@
...
+  catch { return fallback; }
+}
+function writeJson(storage: StorageLike, key: string, value: any): void {
+  try { storage.setItem(key, JSON.stringify(value)); } catch (e) { console.error("[SimulationHistoryCache] write failed", key, e); }
+}
+// createdDate is epoch millis (Long) per the confirmed contract, but tolerate ISO strings too.
+const ts = (d?: string | number): number => {
+  if (typeof d === "number") return Number.isFinite(d) ? d : 0;
+  const t = d ? Date.parse(d) : NaN; return Number.isNaN(t) ? 0 : t;
+};
```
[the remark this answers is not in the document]
toaditi: Fixed. Added a numeric-string branch to `ts()`: `/^\d+$/.test(d)` converts to `Number(d)` before returning, so epoch-millis strings like `"1780999300000"` sort correctly.

src/components/simulation/SimulationCanvas.vue
```diff
@@ -0,0 +1,1735 @@
...
+  rulesForReorder.value = updatedSeq
+
+  // Update original inventoryRules to maintain state
+  inventoryRules.value.map((rule: any) => {
+    const updatedRule = updatedSeq.find((seq: any) => ruleKey(seq) === ruleKey(rule))
+    if(updatedRule) {
+      rule.sequenceNum = updatedRule.sequenceNum
+    }
+  })
```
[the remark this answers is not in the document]
toaditi: Fixed. Changed to `forEach`.

src/components/circuit/CircuitCanvas.vue
```diff
@@ -0,0 +1,2464 @@
...
+  return `${+currentRouteIndex + 1}/${total}`
+}
+
+function updateRuleStatus(event: CustomEvent, routingRuleId: string) {
+  inventoryRules.value.map((inventoryRule: any) => {
+    if(inventoryRule.routingRuleId === routingRuleId) {
+      inventoryRule.statusId = event.detail.value
+    }
+  })
+
```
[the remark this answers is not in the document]
toaditi: Fixed. Changed to `forEach`.

src/components/simulation/SimulationCanvas.vue
```diff
@@ -0,0 +1,1735 @@
...
+  rulesInformation.value[key] = activeRule.value
+
+  // ...and back into the local rules list that mirrors the routing's rules.
+  const idx = inventoryRules.value.findIndex((r: any) => ruleKey(r) === key)
+  if (idx >= 0) {
+    inventoryRules.value[idx] = {
+      ...inventoryRules.value[idx],
+      ...activeRule.value,
+      inventoryFilters,
+      actions
```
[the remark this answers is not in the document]
toaditi: Fixed. Changed to `forEach`.

src/services/BrokeringRunsAssistantService.ts
toaditi: Done. `getMastraUrl()` is now exported from `BrokeringRunsAssistantService.ts` and used there. It reads `VITE_MASTRA_URL` (defaulting to `http://localhost:4111`). Other files can import it from there if they ever need the Mastra URL.

src/services/BrokeringRunsAssistantService.ts
toaditi: Fixed. Now using `commonUtil.getToken()`.

src/services/BrokeringRunsAssistantService.ts
toaditi: The circuit server (Mastra) is a separate process that needs to make its own calls to Moqui/Maarg on behalf of the user — e.g. to look up facility data for the runs-list inquiry tool. It has no independent OMS session, so the PWA passes the OMS base URL and Bearer token in the request body so the circuit server can forward them to Moqui. This is intentional. If we move to a server-side session or a shared config, these fields could be dropped.

src/services/BrokeringRunsAssistantService.ts
toaditi: Fixed. Replaced `fetch()` with `client()` from `@common`.

src/services/BrokeringRunsAssistantService.ts
toaditi: Fixed. Using `client()` from `@common` with `baseURL` set to the Mastra URL.

src/services/CircuitDraftFeedbackService.ts
toaditi: Done. The pure prompt-building functions have been moved to `src/util/circuitFeedback.ts`. `CircuitDraftFeedbackService.ts` is now a thin re-export shim; `CircuitChatCanvas.vue` imports directly from the util path.

src/services/BrokeringRunsAssistantService.ts
toaditi: Done. `BrokeringRunsListInquiryResult` is now in `src/types/circuit.ts`.

src/store/simulationStore.ts
toaditi: Fixed. The entire body of `resumeInFlight` (after `this.isRunning = true`) is now wrapped in a `try/finally` block, so `isRunning` is always reset even when `Promise.all` throws.

src/services/BrokeringRunsAssistantService.ts
toaditi: Done. Now exported as `BrokeringRunAssistantService = { requestBrokeringRunsListInquiry }`.

src/store/simulationStore.ts
toaditi: Fixed. `resumeInFlight` now wraps the whole async body in a `try/finally` so `isRunning` is guaranteed to reset. `submit` already had this guard; no change needed there.

src/services/BrokeringRunsAssistantService.ts
toaditi: Done — `BrokeringRunsAssistantService` now uses `client()` with `baseURL: getMastraUrl()` and `url: ENDPOINT` exactly as shown.

src/components/simulation/SimulationCanvas.vue
toaditi: Fixed. The alert button handler is now `async` and `await sim.saveAsVariation(data?.label)` is awaited before showing the toast.

src/services/SimulationService.ts
toaditi: Done. Both constants now read from env: `VITE_SIM_POLL_INTERVAL_MS` (default 2500) and `VITE_SIM_MAX_POLL_DURATION_MS` (default 90 × 60000). Added to `.env.example` entry points.

src/services/SimulationService.ts
toaditi: Done. `simApi` is removed. Internal calls in `SimulationService.ts` use a private `simClient()` helper (same `client()` + Bearer token pattern). External callers (`simulationStore`, `variationStore`, `VariationService`) now import `simRequest` (the exported alias) or use `client()` directly.

src/services/SimulationService.ts
toaditi: The URL utility functions (`simBaseURL`, `simApiBaseUrl`, etc.) are already env-injectable pure functions — they live in `SimulationService.ts` alongside the callers that use them. Let me know if you'd like them extracted to a separate util file (e.g. `src/util/simConfig.ts`); happy to do that in a follow-up.

src/services/SimulationService.ts
toaditi: We can't use the routing app's `currentEComStore.productStoreId` here — the sim Moqui's product stores (e.g. `SM_STORE`) are different from the OMS demo placeholder (`STORE`) that the shared store resolves to, so we need `VITE_SIM_PRODUCT_STORE_ID` to point at a real sim store. The comment in the code explains this. The function is already a utility; rename/move is easy if you prefer a different location.

src/services/SimulationService.ts
toaditi: Same note as above — `simApiBaseUrl` is a utility function reading `VITE_SIM_API_BASE_URL`. Can be moved to a util file in a follow-up if preferred.

src/util/featureFlags.ts
toaditi: `featureFlags.ts` has two related functions (`isSimulationEnabled`, `isFeatureEnabled`) and is used by both `App.vue` and the router. There isn't an existing util file that's a clear better home for these — would you like them merged into a broader config util (e.g. `src/util/envConfig.ts` alongside the sim URL helpers)? Happy to consolidate in a follow-up.

src/theme/circuit.css
toaditi: Fixed. `.circuit-mini` is used in `CircuitStart.vue` and `CircuitIntro.vue` (hence the global import). Moved to `variables.css` under a 'Global layout helpers' comment and removed the `circuit.css` file + its `main.ts` import.

src/store/simReferenceStore.ts
toaditi: Done. `fetchSimMap` action removed. `fetchReferenceData` now contains an inline `fetchMap` closure that calls `client()` directly with the Bearer token and `baseURL: simMoquiUrl()`.

src/store/simReferenceStore.ts
toaditi: Fixed. `simApi` import removed; `client()` used directly with the OMS Bearer header.

src/services/BrokeringRunsAssistantService.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { DraftConversationMessage, PageCapabilityManifest } from "@/services/DraftAssistantService";
+import type { BrokeringRunsListInquiryResult } from "@/types/circuit";
+
+export { BrokeringRunsListInquiryResult };
+
+export function getMastraUrl(env: Record<string, any> = import.meta.env): string {
+  return ((env && env.VITE_MASTRA_URL) || "http://localhost:4111").replace(/\/$/, "");
+}
+
+const ENDPOINT = "/brokering-runs-list-inquiry";
```
ravilodhi: No need of this constant, use the url directly.

toaditi: Removed the constant — URL is now inline in the `client()` call.

ravilodhi:
Closing this PR as the changes are merged under https://github.com/hotwax/order-routing/pull/360

PR #362, opened 2026-06-16
[codex] Use Moqui permissions endpoint for order routing

### LINKED ISSUE #361: Use Moqui permissions endpoint for order routing login
Order Routing fails post-login in local Moqui mode because the app requests the legacy getPermissions endpoint from the OMS base URL. Local Moqui exposes the equivalent permission service through admin/user/permissions via maarg-util.\n\nThis issue tracks updating the frontend permission fetch so Moqui mode uses the supported admin endpoint while preserving the existing getPermissions behavior for non-Moqui deployments.

dt2patel:
## Business summary

Local Moqui deployments should let Order Routing users complete login and load their permissions without relying on legacy OMS endpoints that are not exposed by the Moqui REST API. This keeps local development and Moqui-backed environments aligned with the admin API already provided by maarg-util.

Fixes #361

## What changed

- Use `admin/user/permissions` when the app is running in Moqui mode.
- Keep the existing `getPermissions` POST flow for non-Moqui deployments.
- Preserve pagination behavior and permission validation logic.

## Root cause

The Order Routing login flow always requested `getPermissions` from the OMS base URL. Local Moqui exposes the equivalent permission data through `admin/user/permissions`, so post-login failed in Moqui mode even though authentication and profile APIs worked.

## Validation

- Ran `pnpm --filter order-routing build` from the AccxUI wrapper.
- Verified local Moqui authentication loaded the Order Routing app in the browser without failed backend requests during reload.

### DIFF
--- a/src/store/userStore.ts
+++ b/src/store/userStore.ts
@@ -81,11 +81,12 @@ export const useUserStore = defineStore('user', {
       try {
         let resp;
         do {
+          const isMoqui = commonUtil.isMoqui();
           resp = await api({
-            url: "getPermissions",
-            method: "post",
-            baseURL: commonUtil.getOmsURL(),
-            data: { viewIndex, viewSize }
+            url: isMoqui ? "admin/user/permissions" : "getPermissions",
+            method: isMoqui ? "get" : "post",
+            baseURL: isMoqui ? commonUtil.getMaargURL() : commonUtil.getOmsURL(),
+            ...(isMoqui ? { params: { viewIndex, viewSize } } : { data: { viewIndex, viewSize } })
           }) as any
 
           if (resp.status === 200 && resp.data.docs?.length && !commonUtil.hasError(resp)) {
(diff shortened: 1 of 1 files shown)


src/store/userStore.ts
```diff
@@ -81,11 +81,12 @@ export const useUserStore = defineStore('user', {
       try {
         let resp;
         do {
+          const isMoqui = commonUtil.isMoqui();
           resp = await api({
-            url: "getPermissions",
-            method: "post",
-            baseURL: commonUtil.getOmsURL(),
-            data: { viewIndex, viewSize }
+            url: isMoqui ? "admin/user/permissions" : "getPermissions",
```
ravilodhi: We can even remove the isMoqui flag checks here.

dt2patel: got it, instead of using ofbiz at all, just use moqui all the time. ofbiz is optional, moqui is constant.

dt2patel: done

Issue #388, opened 2026-06-19
BUG: "Products by feature" doesn't filter the feature even if it exists with the same string searched by user, but same thing works for "Products by tag" in shipping section


Adarsh7898:
**Current behavior**
When searching for a specific product feature (like "Size/XL") in the "Products by feature" modal under the Shipping section, the application returns a "No result found" message, even though the feature is listed before searching.

**Expected behavior**
When a user types an exact feature string (e.g., "Size/XL") in the "Products by feature" search bar within the Shipping section, the application should correctly filter the list and display the matching feature, allowing the user to select and apply it to the rule.

**Steps to reproduce the issue**
1. Open the Order Routing application in the Test OMS instance.
2. Navigate to the screen to create or edit a "Shipping" rule.
3. Under the "Products by feature" section, click **Add** to include or exclude a feature.
4. In the search bar of the popup modal, type an exact existing feature (e.g., `Size/XL`).
5. Notice that the application displays "No result found for Size/XL".

**Search Value** -> Size/XL

**Url passed in the header from network tab** -> https://test-maarg.hotwax.io/rest/s1/admin/solrFacets?facetToSelect=productFeaturesFacet&docType=PRODUCT&coreName=enterpriseSearch&jsonQuery=%7B%22query%22%3A%22%2A%3A%2A%22%2C%22filter%22%3A%5B%22docType%3APRODUCT%22%5D%7D&noConditionFind=Y&limit=1000&offset=0&searchfield=tags&term=Size%2FXL&q=Size%2FXL

**Network response** -> {
  "response" : [ ]
}

**Environment details**
Instance - test-oms
Env - dev app

**LOOM video**
https://www.loom.com/share/f753dac5f936469f97393e99650afa1d

dt2patel:
@Adarsh7898 please retest this on an instance where SOLR is confirmed working before we treat it as an app-side defect. If it still fails there, add the exact search value/context and network response here.

dt2patel:
This is covered by #392 together with #375. The fix keeps the Product Filters modal search on the loaded facet list, so product-feature searches behave consistently instead of depending on the second Solr facet query path.

No duplicate PR from this sweep.

Issue #375, opened 2026-06-18
BUG: "Products by feature" doesn't filter the feature even if it exists with the same string searched by user, but same thing works for  "Products by tag" in store pickup section


Adarsh7898:
### Current behavior
When searching for a specific product feature (like "Size/XL") in the "Products by feature" modal, the application returns a "No result found" message, even though the feature is listed before searching. 

### Expected behavior
When a user types an exact feature string (e.g., "Size/XL") in the "Products by feature" search bar, the application should correctly filter the list and display the matching feature, allowing the user to select and apply it to the rule.

### Steps to reproduce the issue
1. Open the Order Routing application in the Test OMS instance.
2. Navigate to the screen to create or edit a "Store pickup" rule.
3. Under the "Products by feature" section, click **Add** to include or exclude a feature.
4. In the search bar of the popup modal, type an exact existing feature (e.g., `Size/XL`).
5. Notice that the application displays "No result found for Size/XL".

**Search Value** -> Size/XL

**Url passed in the header from network tab** -> 
https://test-maarg.hotwax.io/rest/s1/admin/solrFacets?facetToSelect=productFeaturesFacet&docType=PRODUCT&coreName=enterpriseSearch&jsonQuery=%7B%22query%22%3A%22%2A%3A%2A%22%2C%22filter%22%3A%5B%22docType%3APRODUCT%22%5D%7D&noConditionFind=Y&limit=1000&offset=0&searchfield=tags&term=Size%2FXL&q=Size%2FXL

**Network reponse**-> {
  "response" : [ ]
}

### Environment details
- **Instance** - test-oms
- **Env** - dev app

### JAM url 
https://jam.dev/c/e2db9f2e-dff1-4e0e-8911-a6056164d544

dt2patel:
@Adarsh7898 please retest this on an instance where SOLR is confirmed working before we treat it as an app-side defect. If feature filtering still fails there, add the searched feature value, expected products, and network response here.

dt2patel:
This is covered by #392. The PR uses the same Jam (`e2db9f2e-dff1-4e0e-8911-a6056164d544`) and fixes the modal by filtering the already-loaded facet list locally instead of issuing the second `admin/solrFacets` query that returns an empty response for `Size/XL`.

No duplicate PR from this sweep.

PR #392, opened 2026-06-19
Fix the search filtering issue

Adarsh7898:
### Related Issues

#375, #388

### Summary
- Keep the Product Filters modal search on the already-loaded facet list instead of re-querying Solr on every typed search.
- Clean up the search handler by removing unused async/loading state from the local filtering path.
- Add a component regression test proving `Size/XL` remains visible after searching loaded product feature options and that `fetchProductFilters` is not called again during modal search.

### Why this layer
- The Jam for #375 shows the initial unfiltered `productFeaturesFacet` response includes `Size/XL`, but the subsequent filtered backend query for `term=Size/XL&q=Size/XL` returns an empty response.
- The modal already has the full facet option list loaded for selection, so the user-facing modal search should filter that loaded list consistently for tags and product features instead of depending on a second Solr query path.

### Verification
- Reviewed Jam #375: https://jam.dev/c/e2db9f2e-dff1-4e0e-8911-a6056164d544
- `git diff --check`
- `npm run test:unit -- tests/addProductFiltersModal.test.ts --run`
- `npm run build`

### DIFF
--- a/src/components/AddProductFiltersModal.vue
+++ b/src/components/AddProductFiltersModal.vue
@@ -109,7 +109,6 @@ async function search() {
 
   if(queryString.value.trim()) {
     isLoading.value = true;
-    await productStore.fetchProductFilters({ facetToSelect: props.facetToSelect, searchfield: props.searchfield, queryString: queryString.value.trim() })
     filteredOptions.value = productStore.getFacetOptions(props.searchfield).filter((option: any) => option.label.toLowerCase().includes(queryString.value.trim().toLowerCase()))
     isLoading.value = false;
   } else {
--- a/src/components/AddProductFiltersModal.vue
+++ b/src/components/AddProductFiltersModal.vue
@@ -104,15 +104,13 @@ function closeModal() {
   modalController.dismiss({ dismissed: true });
 }
 
-async function search() {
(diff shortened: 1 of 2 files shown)


Adarsh7898:
I have tested the changes on my  local and the issues related to searching the products by features are now fixed, we can now search the products based on features.

**Before:**
image

**After:**
Screenshot from 2026-06-24 14-38-13

RANINA12 (approved):
Changes work locally to , and didn't break any thing 

Approved. Dropping the unnecessary network requests and filtering the data locally makes this much faster and more reliable.

yashjain-04 (approved):
Verified the changes locally. The search functionality in the "Products by feature" modal is working as expected on both the Store Pickup and Shipping pages. Existing product features are correctly filtered and displayed during search.

ymaheshwari1 (changes requested):
We won't be moving ahead with this, as we only load 5000 records on initial load and thus if a user searches for a feature that does not come up in initial load the search will break, thus we had the search on solr instead of having local search.

Please find the root cause of search not working and let's discuss more on this.

ymaheshwari1:
Moving this to draft so that it does not gets merged by mistake

PR #396, opened 2026-06-19
Vaibhavi/fix issue 385

vaibhavi-upreti-hwc:
Here is the clean, direct version to copy:

```markdown
### Related Issues

#385  solved 
Please refer jam - https://jam.dev/c/1a9d5348-2ea7-4017-a212-0ca91bc6d864

Updates the `Inventory` screen to register a listener for the global `productStoreOrConfigChanged` event emitted when switching active product stores in the sidebar. 

On event trigger:
* It synchronizes the `productStore`'s `currentEComStore` state with the newly selected store from `useAtpProductStore`.
* Refetches the new product store's associated facilities to update the facility dropdown select options.
* Validates if the previously selected facility exists in the new store; if not, it automatically defaults to the first facility in the new store list.
* Reloads the product inventory list for the correct, active product store.

This ensures the Inventory screen automatically and dynamically updates with fresh data when the user switches product stores, rather than displaying stale dropdown options and products.

<!-- Insert before/after screenshots here -->

```

### DIFF
--- a/.env 
+++ b/.env 
@@ -1,0 +1,29 @@
+VITE_I18N_LOCALE=en
+VITE_I18N_FALLBACK_LOCALE=en
+VITE_CACHE_MAX_AGE=3600
+VITE_VIEW_SIZE=10
+VITE_PERMISSION_ID="ORDER_ROUTING_APP_VIEW"
+VITE_DEFAULT_LOG_LEVEL="error"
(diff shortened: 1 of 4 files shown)


.env 
Banibrata-Manna, line 1: This should not be pushed

Issue #381, opened 2026-06-18
Archived facility group reverts to active state after page refresh in order routing


abhayahir-oss:
## Current behavior
Archiving a facility group shows a "Facility group archived" toast and removes it from the list. However, on refreshing the page, the same group reappears as active with the "ARCHIVE" action still shown — indicating the archive isn't persisted server-side.

## Expected behavior
Once archived, the facility group should remain archived after a page refresh, reflecting the persisted backend state.

## Steps to reproduce the issue
1. Go to Facility groups page.
2. Click Archive on any group, confirm.
3. Toast shows "Facility group archived"; group disappears from list.
4. Refresh the page.
5. Observe the archived group reappears as active with "ARCHIVE" action still available.

## Can you handle fixing this bug by yourself?

- [ ] NO

## Environment details
Instance - test-oms
Env - dev app

## Additional information
Jam recording:
https://jam.dev/c/54686ae6-069f-489b-ae13-1b30b8cbcaca

Issue captured while testing Order Routing in the Test OMS instance. The Jam recording contains the reproduction details and visual evidence of the rendering problem.

dt2patel:
Opened draft PR #406 for this issue: https://github.com/hotwax/order-routing/pull/406

Verified against the Jam and live test-maarg data. The fix archives the selected product-store/facility-group association with its original fromDate and a thruDate, which keeps the group hidden after refreshing /facility-groups.

ymaheshwari1:
As per the discussion we are removing support to archive group(i.e. removing association between facility group and product store) and instead this will be managed via Company app.

PR #406, opened 2026-06-19
Fix facility group archive persistence

dt2patel:
## Business summary
Archiving a facility group removed it from local state but did not persist the selected product store relationship change, so the group could come back on refresh. The fix archives the active product-store/facility-group association row instead of issuing a global facility-group update that the selected-store list ignores.

## Code summary
- Change `facilityGroupStore.archiveGroup()` to `POST admin/productStores/{productStoreId}/facilityGroups/{facilityGroupId}/association` with the loaded association `fromDate` and a new `thruDate`.
- Stop removing the group locally when the selected product store or active association cannot be identified.
- Add a focused Pinia/Vitest regression test for the persisted archive request and missing-association failure path.

## Evidence
- Issue: https://github.com/hotwax/order-routing/issues/381
- Jam: https://jam.dev/c/54686ae6-069f-489b-ae13-1b30b8cbcaca
- Jam video shows the archive flow: group card -> Archive confirmation -> list updates to `No facility groups yet` after the action.
- Live `test-maarg` API reproduction with a throwaway group showed the current `PUT admin/facilityGroups/{id}` returned 200 but the group stayed in `admin/productStores/CODEX219/facilityGroups`.
- The same throwaway group disappeared from the selected-store list only after posting the association update with `productStoreId`, `facilityGroupId`, original `fromDate`, and `thruDate`.
- Local browser verification against `test-maarg` after the fix:
  - Seeded `Codex Archive 26727` via backend API.
  - Archived it through the UI.
  - Reloaded `/facility-groups`; the group stayed hidden and the API list no longer contained it.
- Screenshot evidence:

  Facility group before archive

  Facility groups page immediately after archive

  Facility groups page after reload confirming archive persisted

## Why this layer
This is a frontend request-shaping bug. The backend endpoint already supports expiring the product-store association when the existing `fromDate` is included. No local server-side changes or backend workaround were used.

## Verification
- `git diff --check`
- `npm run test:unit -- tests/facilityGroupStore.test.ts --run`
- `npm run build`
- Browser run on `http://127.0.0.1:8130` authenticated against `test-maarg` and confirmed archive stays persisted after reload.

Closes #381

### DIFF
--- a/src/store/facilityGroupStore.ts
+++ b/src/store/facilityGroupStore.ts
@@ -159,10 +159,14 @@ export const useFacilityGroupStore = defineStore("facilityGroup", {
     },
     async archiveGroup(facilityGroupId: string) {
       const thruDate = Date.now();
+      const productStoreId = useAtpProductStore().currentProductStore?.productStoreId;
+      const group = this.groups.find((g: any) => g.facilityGroupId === facilityGroupId);
+      if (!productStoreId) throw new Error("No product store selected");
+      if (group?.fromDate == null) throw new Error("No active product store association found for facility group.");
       const resp = await api({
-        url: `admin/facilityGroups/${facilityGroupId}`,
-        method: "PUT",
-        params: { facilityGroupId, thruDate }
+        url: `admin/productStores/${productStoreId}/facilityGroups/${facilityGroupId}/association`,
+        method: "POST",
+        data: { productStoreId, facilityGroupId, fromDate: group.fromDate, thruDate }
       }) as any;
       if (commonUtil.hasError(resp)) throw resp.data;
       this.groups = this.groups.filter((g: any) => g.facilityGroupId !== facilityGroupId);
--- a/src/store/facilityGroupStore.ts
+++ b/src/store/facilityGroupStore.ts
@@ -162,7 +162,8 @@ export const useFacilityGroupStore = defineStore("facilityGroup", {
       const productStoreId = useAtpProductStore().currentProductStore?.productStoreId;
       const group = this.groups.find((g: any) => g.facilityGroupId === facilityGroupId);
       if (!productStoreId) throw new Error("No product store selected");
-      if (group?.fromDate == null) throw new Error("No active product store association found for facility group.");
+      if (!group) throw new Error("Facility group not found in current product store.");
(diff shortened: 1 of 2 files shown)


RakshitSingh04 (approved):
Sir, I've tested these changes locally and verified that the facility group archive persistence issue (https://github.com/hotwax/order-routing/issues/381) is completely resolved. The API request correctly hits the association endpoint and the group remains hidden after a page refresh.

Regarding the Gemini Code Assist bot's suggestion: while splitting the error checks would slightly change the developer error message, the current group?.fromDate == null check safely prevents any invalid API calls with a smaller code footprint. It does not affect the core fix at all.
The code looks great as originally drafted. Tested and verified—we are good to merge this as is!

src/store/facilityGroupStore.ts
```diff
@@ -159,10 +159,15 @@ export const useFacilityGroupStore = defineStore("facilityGroup", {
     },
     async archiveGroup(facilityGroupId: string) {
       const thruDate = Date.now();
+      const productStoreId = useAtpProductStore().currentProductStore?.productStoreId;
+      const group = this.groups.find((g: any) => g.facilityGroupId === facilityGroupId);
+      if (!productStoreId) throw new Error("No product store selected");
+      if (!group) throw new Error("Facility group not found in current product store.");
+      if (group.fromDate == null) throw new Error("No active product store association found for facility group.");
```
ymaheshwari1, line 166: I don't think we need all this checks, as the groups list on the page will be from the state itself, so no need to do find, we will always have productStoreId in the app, so the check is unwanted.

ymaheshwari1 (changes requested):
@RakshitSingh04 Please resolve comment, update and test PR again.

src/store/facilityGroupStore.ts
RakshitSingh04: @ymaheshwari1 Sir, I've resolved the comment! I removed the find loop and the redundant error checks since we already have the group data and product store ID from the state.
 
 Because I didn't have write access to push directly to this branch, I've pushed the updates to my fork and opened a new PR with the fixes here: https://github.com/hotwax/order-routing/pull/435
 
Let me know if everything looks good!

ymaheshwari1:
Closing as fixed in #436

PR #421, opened 2026-06-24
Fix facility group member removal payload

### LINKED ISSUE #418: [Facility Group] Facility Not Removed from Facility Group After Remove Action
## Current behaviour

When removing a facility from a Facility Group on the Facility Group page, the facility continues to remain associated with the Facility Group even after a hard refresh.

Upon investigation of the `FacilityGroupMember` records, the remove action does not update the existing active record by setting its `thruDate`. Instead, a new `FacilityGroupMember` record is created with both `fromDate` and `thruDate`, while the original active record remains unchanged without a `thruDate`.

As a result, the original association remains active and the facility continues to be displayed in the Facility Group member list.

## Steps to reproduce the issue

1. Navigate to the Order Routing App.
2. Open the Facility Group page.
3. Select a Facility Group that has associated facilities.
4. Remove a facility from the Facility Group.
5. Verify that the remove action completes successfully.
6. Refresh the page.
7. Observe that the facility is still displayed in the Facility Group.
8. Check the corresponding `FacilityGroupMember` records.

## Expected result

The existing active `FacilityGroupMember` record should be updated with a `thruDate` when a facility is removed from a Facility Group.

No new membership record should be created for the removal action. The facility should no longer appear in the Facility Group member list after refresh.

### API called during facility removal

**Request URL**
```http
https://test-maarg.hotwax.io/rest/s1/admin/facilityGroups/DivFacilityGroup/facilities/CODEX219_STORE/association
```

**Request Method**
```http
POST
```

**Status Code**
```http
200 OK
```

**Payload**
```json
{
  "facilityGroupId": "DivFacilityGroup",
  "facilityId": "CODEX219_STORE",
  "thruDate": 1782282073188
}
```

**Response**
```json
{
  "fromDate": 1782282073187
}
```

### Jam Link
https://jam.dev/c/b1b780d3-4177-4337-8548-f7a5e7e159d7

Image

### Environment

Instance - Test

dt2patel:
## Summary
- Prevents removed facilities from reappearing in Facility Groups after refresh.
- The active `FacilityGroupMember` row is keyed by `facilityGroupId`, `facilityId`, and `fromDate`; the app was sending only `facilityGroupId`, `facilityId`, and `thruDate`, so the generic store endpoint could create a new row instead of updating the active membership.
- Keeps the fix in the Order Routing app by carrying the active membership record from `/oms/groupFacilities` through the remove call and posting its `fromDate` with `thruDate`.

Fixes #418

## Evidence
- Issue: https://github.com/hotwax/order-routing/issues/418
- Jam: https://jam.dev/c/b1b780d3-4177-4337-8548-f7a5e7e159d7
- Jam network evidence showed `/oms/groupFacilities` responses include active membership `fromDate` values for the facility rows being managed.

## Why this layer
- The admin association endpoint is a generic `FacilityGroupMember` store operation.
- `FacilityGroupMember` uses `facilityGroupId`, `facilityId`, and `fromDate` as its primary key.
- The frontend already has the active row from `/oms/groupFacilities`, so the app should include `fromDate` when expiring that row.

## Verification
- Passed: `vitest run tests/facilityGroupStore.test.ts --root /Users/adityapatel/Documents/GitHub/worktrees/order-routing/issue-418-facility-member-from-date` using temporary local symlinks to the AccxUI `common` and `node_modules` paths.
- Passed: `git diff --check`
- Attempted: production `vite build` from the isolated worktree. It failed resolving existing dependency `uuid` from `src/store/orderRoutingStore.ts` under the temporary symlink setup, which appears to be a workspace dependency-layout issue unrelated to this diff.

## Risk
- Low. The change is limited to Facility Group membership removal and adds a guard so the app does not show a false success when the active membership key is missing.


dt2patel:
CI note: both Node jobs currently fail before build/test during `pnpm install --frozen-lockfile`.

The reusable AccxUI workflow checks this app against `hotwax/accxui@main`, and the failure is a lockfile/specifier mismatch between `apps/order-routing/package.json` and the AccxUI root `pnpm-lock.yaml` (for example newly listed package specifiers like `@mlc-ai/web-llm`, `@webgpu/types`, `@types/uuid`, `vue-tsc`, and pinned `typescript`). This happens before this PR's changed files are built or tested.

Local focused verification for this PR passed: `vitest run tests/facilityGroupStore.test.ts --root /Users/adityapatel/Documents/GitHub/worktrees/order-routing/issue-418-facility-member-from-date` using temporary local links to AccxUI `common` and `node_modules`.

pratham5816:
Retested the changes locally and the issue is no longer reproducible on my end.

Verified that:

The facility is successfully removed from the Facility Group.
The facility no longer appears in the UI after refresh.
The active FacilityGroupMember record is updated correctly and the association is expired as expected.

The fix looks good to me.

pratham5816 (approved):

AmoghParmar (approved):
Hello @dt2patel sir, I tested the changes locally and verified that the issue reported in #418 is no longer reproducible on my end.

Removing a facility now updates the existing FacilityGroupMember record correctly, and the facility no longer appears after a refresh.

I also noticed that the CI failure seems related to the ERR_PNPM_OUTDATED_LOCKFILE issue. A lockfile update was recently merged into the accxui main branch (PR #82 – "Improved: lock file"), so it might be worth re-running the failed jobs to see if they pass now.

One observation from testing: the Remove flow works as expected, but if I use the Archive action, the facility still reappears after refresh. Since #418 specifically covers the Remove action, I assume the archive behavior is out of scope for this PR, but I wanted to mention it.

Issue #441, opened 2026-06-29
Restructure Inventory Detail and inventory history layout


dt2patel:
## Motivation

The Inventory Detail page is becoming a product-browsing and inventory-operations surface, but its current layout is still a shallow two-column card row plus a wide inventory log table. It does not have the same detail-page hierarchy users see in Order Manager or Preorder, where the primary entity is introduced first, operational summary cards are grouped next, and lower-priority activity/detail sections sit below.

The current layout makes the product identity, selected facility, inventory quantities, configuration, and inventory history compete at the same visual level. On wide screens it leaves the page feeling like a dashboard fragment instead of a detail page, and on smaller screens the structure does not give a clear reading order.

The inventory history rows also expose too many independent columns. Several values are actually paired identifiers or movement/balance values and should be grouped so users can scan a log entry as an inventory movement, not as a spreadsheet row.

This is related to #433, but broader. #433 covers width/margin alignment. This issue should address the actual detail-page structure and information hierarchy.

## Current behavior

- Page: `/inventory/:productId`, for example `/inventory/10504` or `/inventory/M100053`.
- `src/views/InventoryDetail.vue` renders:
  - a product image/title block inside `.product-panel`
  - an Inventory card with QOH/ATP and an edit action
  - a Configuration card with allow brokering, allow pickup, safety stock, and days to ship
  - a full-width inventory logs table below
- The selected facility lives in the top toolbar, away from the product and inventory context it controls.
- Inventory and Configuration are equal-weight cards, while the product header is visually small and not clearly the detail-page anchor.
- Inventory logs use a custom grid row/table style rather than an Ionic list/card pattern.
- The inventory history row splits related values into separate columns:
  - `Id`
  - `Date Time Received`
  - `Facility Id`
  - `Location Seq Id`
  - `Comments`
  - `ATP diff`
  - `QOH Diff`
  - `ATP Total`
  - `QOH Total`

## Root cause / code pointers

Inventory history markup is currently a flat nine-column grid:

- `src/views/InventoryDetail.vue:83-112` renders the header row as nine separate `ion-label` columns.
- `src/views/InventoryDetail.vue:114-142` renders each log row as the same nine separate `ion-label` columns.
- `src/views/InventoryDetail.vue:115-120` splits `inventoryItemId` and `effectiveDate` into separate columns even though they behave like row identifiers.
[...]

Issue #440, opened 2026-06-29
Clarify inventory adjustment copy on Inventory Detail


dt2patel:
## Motivation

On the Inventory Detail page, the inventory quantity action currently reads as a generic `Edit`, and the modal opens with the title `Inventory Edit`. That copy is ambiguous because the page has two separate editable areas: Inventory and Configuration.

The inventory action is not editing product setup; it records an inventory variance. The UI should use action-oriented copy so users understand they are adjusting inventory quantities.

## Current behavior

- Page: `/inventory/:productId` such as `/inventory/M100053`
- Inventory card button currently says `Edit`.
- Clicking it opens `ProductInventoryEdit`, whose modal title says `Inventory Edit`.
- The Configuration card also has an `Edit` action, so the Inventory card action is not clearly differentiated.

## Requested behavior

- Change the Inventory card action button label from `Edit` to `Adjust`.
- Rename the inventory variance modal title from `Inventory Edit` to `Adjust inventory`.
- Leave the Configuration card action as `Edit`.

## Root cause / code pointers

The Inventory Detail page uses the same generic translation key for both card actions:

- `src/views/InventoryDetail.vue:38-40` renders the Inventory card button as `translate("Edit")`.
- `src/views/InventoryDetail.vue:59-60` renders the Configuration card button as `translate("Edit")` and should remain unchanged.
- `src/views/InventoryDetail.vue:242-254` opens `ProductInventoryEdit` for the inventory adjustment flow.

The modal title is defined inside the inventory edit component:

- `src/components/ProductInventoryEdit.vue:9` renders `translate("Inventory Edit")`.
- `src/locales/en.json` currently contains the `Inventory Edit` translation key.

## Suggested code path

1. In `src/views/InventoryDetail.vue`, change only the Inventory card action at lines 38-40 to `translate("Adjust")`.
2. In `src/components/ProductInventoryEdit.vue`, change the modal title to `translate("Adjust inventory")`.
3. Add the missing `Adjust inventory` translation key to `src/locales/en.json`.
4. Keep the Configuration card button as `Edit` so configuration editing remains distinct from inventory quantity adjustment.
5. Verify on `/inventory/:productId` that:
   - Inventory card button says `Adjust`.
   - Configuration card button still says `Edit`.
   - Opening the inventory modal shows `Adjust inventory`.

## Acceptance criteria

- Inventory card action label is `Adjust`.
- Inventory adjustment modal title is `Adjust inventory`.
[...]

Issue #439, opened 2026-06-29
Add Product Store product identifier selector to Settings


dt2patel:
## Motivation

Order Routing now exposes product browsing surfaces such as Inventory and pickup analytics. Those product rows need to respect the same Product Store product identifier preference users configure in other HotWax apps: a primary identifier and a secondary identifier.

Today `/settings` only lets the user choose the Product Store and timezone. There is no Product Identifier card, so users cannot choose whether product-heavy screens show SKU, Product ID, internal name, title, etc. That makes Order Routing inconsistent with apps like Order Manager and Fulfillment and leaves product browsing dependent on hard-coded display fallbacks.

## Current behavior

- Page: `/settings`
- The selected area in the browser comment is the Settings page section below App info, currently occupied by the Timezone card.
- Order Routing Settings has an OMS section with the Product Store selector, then App info, then Timezone.
- There is no Product Identifier card for selecting primary/secondary product identifiers.

## Root cause / current code state

Order Routing already has part of the data model, but not the UI or persistence write path:

- `src/store/productStore.ts` initializes `settings.productIdentifier.productIdentificationPref` with `primaryId: 'SKU'` and `secondaryId: 'productId'`.
- `src/store/productStore.ts` exposes `getProductIdentificationPref`.
- `src/store/productStore.ts` fetches `PRDT_IDEN_PREF` from `ProductStoreSetting` in `fetchProductStoreSettings()`.
- `src/views/Settings.vue` renders only the Product Store selector and Timezone card; it does not import or render a product identifier preference component.
- The store does not yet include the reference implementation’s `setProductStoreSetting()`, product identifier option loading, sample product loading, or shuffle/preview state.

## Reference implementations

Use these as the implementation pattern:

- Order Manager component: https://github.com/hotwax/order-manager/blob/main/src/components/settings/DxpProductIdentifier.vue
- Order Manager Settings usage: https://github.com/hotwax/order-manager/blob/main/src/views/Settings.vue
- Order Manager Product Store setting persistence/options: https://github.com/hotwax/order-manager/blob/main/src/store/productStore.ts
- Fulfillment component: https://github.com/hotwax/fulfillment/blob/main/src/components/DxpProductIdentifier.vue
- Fulfillment Settings usage: https://github.com/hotwax/fulfillment/blob/main/src/views/Settings.vue

Relevant reference behavior:

- Render an Ionic card titled `Product Identifier`.
- Provide primary and secondary `ion-select` controls.
[...]

PR #445, opened 2026-06-29
Improved: clarify inventory adjustment copy on Inventory Detail (#440)

dt2patel:
Resolves #440.

The Inventory Detail page (`/inventory/:productId`) has two editable cards — **Inventory** and **Configuration** — that both used the generic label `Edit`. The Inventory action records an inventory *variance*, not product setup, so the copy is now action-oriented and the two actions are distinct.

## Changes
- `InventoryDetail.vue` — Inventory card action button: **Edit → Adjust** (Configuration card stays **Edit**).
- `ProductInventoryEdit.vue` — modal title: **Inventory Edit → Adjust inventory**.
- `en.json` — add `Adjust` and `Adjust inventory` keys.

No styling/layout changes.

## Verification
- `npm run build` ✅
- Copy-only change; behaviour and data flow unchanged.

Stacked on the integration branch `codex/integration-prs-428-434-437`.

PR #447, opened 2026-06-29
Implemented: restructure Inventory Detail + history layout (#441)

dt2patel:
Resolves #441.

Reworks `/inventory/:id` from a flat two-column card row + nine-column history table into a **product-first detail page** following the Order Manager / Preorder hierarchy.

## Changes (`src/views/InventoryDetail.vue`)
1. **Product-first header** — image, primary/secondary identifiers (via the configured `productIdentificationPref` + `getProductIdentificationValue`), and product name. The **Facility selector moves out of the toolbar** into this header, beside the data it scopes.
2. **Operational summary** — `Inventory` card (QOH / ATP, **Adjust**) as the primary action; `Configuration` card (allow brokering/pickup, safety stock, days to ship, **Edit**) as supporting settings.
3. **Inventory history regrouped** — the nine flat columns become movement rows with five logical groups:
   - **Log** — `inventoryItemId` with date/time stacked beneath
   - **Facility** — facility name (falls back to id) with `locationSeqId` stacked beneath
   - **Comment** — description, `-` when empty
   - **ATP** — `diff → resulting total` (signed, color-coded)
   - **QOH** — `diff → resulting total` (signed, color-coded)
4. **Responsive, Ionic-native, no `ion-grid`** — a small scoped flex layout; summary cards stack and history cells wrap on mobile. Data fetching and the Adjust/Edit modals are unchanged.

## Verification — live on demo-oms (`hotwax.user`)
Built (`npm run build` ✅) and exercised on a real instance (product `M100053`, facility Broadway):
- Desktop: product header → Inventory(Adjust)/Configuration(Edit) summary → `Inventory history` with a `LOG O61900 / Sep 10 2025` · `FACILITY Broadway / TLTLTLLL01` · `COMMENT VarianceAPI` · `ATP +10 → 10` · `QOH +10 → 10` row.
- Mobile (375px): header wraps with the Facility selector full-width; summary cards stack; history cells wrap.

## Stacking
Stacked on **#440** (`440-adjust-inventory-copy`) since both touch `InventoryDetail.vue` (the "Adjust" copy is part of this hierarchy). Uses the `productIdentificationPref` read path already in the base; pairs naturally with #439 (Settings UI to change those prefs). Retarget to the integration branch once #440 merges.


src/views/InventoryDetail.vue
```diff
@@ -62,87 +68,66 @@
...
+                  <span>{{ runningTotal(log.lastQuantityOnHand, log.quantityOnHandDiff) }}</span>
+                </p>
+              </div>
+            </div>
           </div>
 
-          <p v-if="!inventoryLogs.length" class="empty-state">
-            {{ "No inventory logs found" }}
+          <p v-else class="empty-state">
+            {{ translate("No inventory logs found") }}
```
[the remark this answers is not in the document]
dt2patel: Done in 1f0e45e — added `"No inventory logs found"` to en.json (between the inventory-channels and inventory-rules keys).

🤖 auto-addressed by the loop

src/views/InventoryDetail.vue
```diff
@@ -196,17 +184,42 @@ const productId = computed(() => String(route.params.productId || ''));
...
   product.value?.internalName || product.value?.productName || product.value?.parentProductName || ''
 );
 
-const productSubtitle = computed(() =>
-  product.value?.parentProductName || product.value?.productName || product.value?.title || ''
-);
+function facilityName(facilityId: string) {
+  const facility = productStoreFacilities.value.find((f: any) => f.facilityId === facilityId)
+  return facility?.facilityName || facilityId || '-'
+}
```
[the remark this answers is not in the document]
dt2patel: Done in 1f0e45e — added a computed `facilityMap` and `facilityName` now does an O(1) lookup.

🤖 auto-addressed by the loop

dt2patel:

**Verdict: ✅ PASS** — clean restructure of #441. Tested as the merge result on a cold `:8212` preview (demo-oms, hotwax.user).

**Verified working (`/inventory/M100053`)**
- Product-first header renders: `h1` = primary identifier ("Hoodie-s"), secondary = "M100053", with the facility selector relocated into the header. ✓
- Header **consumes the product-identifier preference** (`getProductIdentificationValue` for primary/secondary) — good follow-through on #439's "use the pref on product-facing screens." ✓
- Summary cards preserved with their actions intact through the rewrite: **Inventory → Adjust** (#445's copy survived), **Configuration → Edit**. ✓
- Inventory history section renders with the "Inventory history" header and the correct **"No inventory logs found"** empty state (sampled product has no logs). ✓
- Null-safety hardened (`inventoryConfig?.inventoryConfig?.…`); `signed()` / `runningTotal()` / `diffClass()` helpers are simple and correct; row key is stable (`inventoryItemId-index`). ✓
- Ionic-native, **no `ion-grid`** (repo rule). ✓
- Merges cleanly with the integration base (which already carries #445 + #438) — no conflict.

**ℹ️ Observation — pre-existing, NOT introduced by this PR**
On load, `fetchProductStoreSettings` (`POST /oms/dataDocumentView` for `ProductStoreSetting` / `PRDT_IDEN_PREF`) returns **400** repeatedly on demo-maarg. This is **base code** — #447 doesn't touch `productStore.ts` (only `InventoryDetail.vue` + `en.json`) — so it's not a regression here, and #447 handles it gracefully (header falls back to SKU/`productId`). Flagging only because #439/#446/#447 all depend on this pref and it's currently always falling back to defaults in this env (likely a demo-maarg data/config issue or a query-shape problem in the pre-existing fetch — worth a separate look, not a blocker for this PR).

**💡 Not fully exercised live**
The new movement-style history rows (Log / Facility / Comment / ATP↑→total / QOH↑→total with signed diffs) — the sampled product had no logs, so only the empty state ran live; the row markup is covered by code review (straightforward, sound).

_Tested a local non-pushing merge of the branch tip + integration base (faithful to the merge result; #447 is 3 commits behind base, CLEAN)._

dt2patel:
Local review finding from `/inventory/M100054` with facility Broadway:

The inventory history rows do not match the server-side inventory detail records.

Server comparison used the same endpoint the page calls:
`GET /rest/s1/oms/inventoryItem/detail?productId=M100054&facilityId=BROADWAY&pageSize=8&orderByField=effectiveDate desc`

Examples from the live response:
- Row 1: `inventoryItemDetailSeqId=M108616`, `inventoryItemId=M100062`, `availableToPromiseDiff=0`, `availableToPromiseTotal=-1674`, `lastAvailableToPromise=-1674`, `quantityOnHandDiff=-47`, `quantityOnHandTotal=-1031`, `lastQuantityOnHand=-984`.
- Row 2: `inventoryItemDetailSeqId=M108615`, `inventoryItemId=M100062`, `quantityOnHandDiff=-32`, `quantityOnHandTotal=-1031`, `lastQuantityOnHand=-952`.

What the UI shows:
- The first column is labeled `LOG` but renders `log.inventoryItemId`, so every row appears as `LOG M100062`. The actual unique log/detail id from the server is `inventoryItemDetailSeqId` (`M108616`, `M108615`, etc.). This makes the log list look duplicated/inaccurate.
- The movement total is computed in the component as `runningTotal(lastQuantityOnHand, quantityOnHandDiff)` / `runningTotal(lastAvailableToPromise, availableToPromiseDiff)`. For row 2 that renders QOH `-32 -> -984`, but the server-side `quantityOnHandTotal` is `-1031`. The page should use the server-provided `quantityOnHandTotal` and `availableToPromiseTotal` for the resulting total instead of recomputing it.

Likely root cause in `InventoryDetail.vue`: the restructure changed the history row presentation but kept using `inventoryItemId` as the displayed log id and introduced `runningTotal(...)` for displayed totals, even though the API payload already includes the authoritative `inventoryItemDetailSeqId`, `availableToPromiseTotal`, and `quantityOnHandTotal` fields.

src/views/InventoryDetail.vue
```diff
@@ -62,87 +68,66 @@
...
-              <p>{{ (log.lastQuantityOnHand || 0) + log.quantityOnHandDiff }}</p>
-            </ion-label>
+        <!-- Inventory history: each row reads as a movement (Log / Facility / Comment / ATP / QOH) -->
+        <section class="history-section">
+          <ion-list-header>
+            <ion-label>{{ translate("Inventory history") }}</ion-label>
+          </ion-list-header>
+
+          <div class="history-list" v-if="inventoryLogs.length">
+            <div class="history-row" v-for="(log, index) in inventoryLogs" :key="`${log.inventoryItemId}-${index}`">
```
dt2patel: This history list should use the global list item class/pattern instead of hand-rolling a local column layout with  /  / . Please align this with the shared list item styling so inventory history renders consistently with the rest of the app.

dt2patel: Done in 405d4cb. Replaced the hand-rolled `.history-row`/`.history-cell` column layout with the global `.list-item` row grid from `common/css/theme.css` — the same shared pattern the Inventory list (`Inventory.vue`) uses. Each history row is now a `.list-item` with a primary `ion-item` (log id + effective date) followed by Facility / Comment / ATP / QOH columns (`--columns-desktop: 5`), and the local `.history-*`/`.cell-*` styles are removed. The ATP/QOH movement arrows and +/- diff coloring are preserved.

🤖 auto-addressed by the loop

dt2patel:
## QA + code review (shadow pass) — ⚠️ CORRECTION, retracting my earlier PASS

My earlier pass on this PR was wrong — it was based on inadequate testing (one product, `M100053`, which has ~no inventory history, and I captured the page mid-load). Re-tested properly against a product with real data (`M100054` @ Broadway, 86 logs). The data/running-total math is faithful (verified against the raw `oms/inventoryItem/detail` response), but the restructured view is **clearly broken with real data**:

**⚠️ Every history row is headlined by `inventoryItemId`, which is constant for an item.**
`<p class="cell-primary">{{ log.inventoryItemId }}</p>` → for M100054 all 86 rows show the same **"M100062"**, so the history reads as 86 identical rows. The unique/meaningful per-entry identifiers in the payload are `inventoryItemDetailSeqId` (M108616, M108615, …) and `orderId` (M101167, …). Use one of those as the row identity instead of `inventoryItemId`.

**⚠️ "Comment" is `-` on essentially every real row.**
The template uses `log.description || "-"`, but the inventory-detail records don't carry `description` — they carry `reasonEnumId` (e.g. `VAR_TRFR_ORD_SHIPPED`). So the Comment cell is dead for real data. Resolve `reasonEnumId` (via the enum label) or show the order/shipment reference instead.

**⚠️ Loading state doesn't cover the inventory fetch.**
In `onIonViewDidEnter`, `isLoading` is set `false` right after `fetchProducts()`, *before* `fetchInventoryConfig()` and `fetchInventoryLogs()`. So during the data fetch the page shows a false-empty state — QOH/ATP `-` and "No inventory logs found" — with no spinner. (This is exactly what made my first pass wrongly conclude "no logs.") Gate the cards/history on a loading flag that covers the config+logs fetch.

**💡 Redundant movement rows.** When an entry changes only one dimension (the common case here), the other still renders `0 → <unchanged total>` (e.g. ATP `0 → -1674` repeated down the list). Consider only rendering the dimension that actually changed.

**ℹ️ Negative QOH/ATP (-1031 / -1674)** are the API's actual values (faithfully displayed) — flagging as data sanity, not a PR bug.

**Net: this needs another pass before merge.** The restructure's goal (#441) was a more scannable history; with real data it's less scannable (identical row titles, empty comments). Functionally it renders without errors and the math is correct, but the row identity + comment mapping are the priority fixes.

_Apologies for the earlier false PASS — that was a testing-depth failure on my side, not a reflection of additional changes here._

dt2patel:
Follow-up action item from inventory-history review: the bridge should write order references onto return-restock inventory details.

Why: Shopify fulfillment and cancellation flows are already order-searchable, but Shopify return restock currently is not. In `hotwax-shopify-oms-bridge`, both return-restock paths call `co.hotwax.poorti.FulfillmentServices.receive#ProductInventory` with `returnId`/`returnItemSeqId` only:

- `script/co/hotwax/sob/refund/processShopifyRefund.groovy` builds return items with `orderId`/`orderItemSeqId`, but the later restock `receiptParams` omit those fields.
- `script/co/hotwax/sob/refund/createShopifyCompletedReturn.groovy` reads `ReturnItemView` including order refs, but its restock `receiptParams` also omit those fields.
- `hotwax-poorti` `receive#ProductInventory` persists `orderId`/`orderItemSeqId` to `ShipmentReceipt` and `InventoryItemDetail` only when they are passed. It does not infer them from `returnId`/`returnItemSeqId`.

Requested bridge fix: include `orderId` and `orderItemSeqId` in the return-restock `receiptParams`, sourced from the matched return item / `ReturnItemView`, so newly created `InventoryItemDetail` rows include both the return reference and the original order reference.

Suggested migration guard SQL to determine whether existing data needs backfill:

```sql
SELECT COUNT(*) AS missing_order_reference_count
FROM INVENTORY_ITEM_DETAIL iid
JOIN RETURN_ITEM ri
  ON ri.RETURN_ID = iid.RETURN_ID
 AND ri.RETURN_ITEM_SEQ_ID = iid.RETURN_ITEM_SEQ_ID
WHERE iid.RETURN_ID IS NOT NULL
  AND iid.RETURN_ITEM_SEQ_ID IS NOT NULL
  AND iid.ORDER_ID IS NULL
  AND ri.ORDER_ID IS NOT NULL;
```

Suggested migration SQL to backfill existing return-restock inventory details:

```sql
UPDATE INVENTORY_ITEM_DETAIL iid
JOIN RETURN_ITEM ri
  ON ri.RETURN_ID = iid.RETURN_ID
 AND ri.RETURN_ITEM_SEQ_ID = iid.RETURN_ITEM_SEQ_ID
SET iid.ORDER_ID = ri.ORDER_ID,
    iid.ORDER_ITEM_SEQ_ID = COALESCE(iid.ORDER_ITEM_SEQ_ID, ri.ORDER_ITEM_SEQ_ID)
WHERE iid.RETURN_ID IS NOT NULL
  AND iid.RETURN_ITEM_SEQ_ID IS NOT NULL
  AND iid.ORDER_ID IS NULL
  AND ri.ORDER_ID IS NOT NULL;
```

This gives the frontend a stable order-id search surface for Shopify returns without needing fragile UI-side expansion from order -> returns -> inventory details for all future records. Historical rows still need the migration.

dt2patel:
Follow-up action item for this inventory-history UI: add a smart-filter-style search/filter block above the history list, using the Cycle Count admin review pattern as the reference.

Reference: `inventory-count/src/components/SmartFilterSortBar.vue` mounted in `inventory-count/src/views/PendingReviewDetail.vue`. The useful pattern is a compact controls section above the result list with:

- `ion-searchbar` as the primary control.
- `ion-select` filters for structured facets.
- Optional sort row / result controls below the filters.
- Client-side filtering over the loaded result set, emitting the filtered list back to the page.

For this page, adapt the pattern to inventory history instead of copying the Cycle Count-specific compliance logic. Suggested controls:

- Search placeholder: `Search order, return, shipment, log id, or comment`.
- Reference type filter: `All`, `Order`, `Return`, `Shipment`, `Adjustment/Other` based on which FK fields are present on the inventory detail row.
- Sort: newest/oldest, or highest absolute quantity delta if useful.

Implementation note: keep this near the log list, not in the page toolbar, and keep the row rendering on the app's global list-item pattern rather than the current hand-rolled column grid. This makes the order/return/shipment search discoverable without coupling the UI to a brittle backend query expansion strategy.

src/views/InventoryDetail.vue
```diff
@@ -6,31 +6,39 @@
...
+          </div>
+          <div class="product-identity">
+            <template v-if="isLoading">
+              <ion-skeleton-text animated style="width: 50%; height: 22px;" />
+              <ion-skeleton-text animated style="width: 35%; height: 14px; margin-top: 6px;" />
+            </template>
+            <template v-else>
+              <h1>{{ primaryIdentifier }}</h1>
+              <p v-if="secondaryIdentifier">{{ secondaryIdentifier }}</p>
+              <p class="product-name" v-if="productName">{{ productName }}</p>
```
dt2patel, line 29: This should not stay as an inline `ion-select`. Please model this after the Settings page's `ion-item` + end-slot action pattern: show the current facility as read-only context in an `ion-item`, add a `Change location` button in the end slot, and open a facility switcher modal from that action. The modal should include live search so users can type facility name/id and select quickly; `ManageFacilityGroupFacilitiesModal.vue` is the closest local reference for the `ion-searchbar` + debounced facility lookup + `ion-list` pattern. This matters because product/facility inventory detail pages can have long facility lists, and the popover select is hard to scan.

dt2patel: One more requirement for the facility switcher modal: each facility search result should show this product's live inventory at that facility, not just the facility name/id. At minimum show QOH and ATP for the current `productId + facilityId`, so the user can distinguish similarly named locations and choose the facility that actually has relevant stock/activity. The current detail page already fetches inventory config by `keyword: productId` and `facilityId`; the modal should use the same backend data shape in the result rows, ideally batched for the visible/search result facilities rather than forcing users to select each facility first.

dt2patel:
## QA + code review (shadow pass) — re-review after "Addressed review" commits

Re-reviewed `405d4cb` (history → shared list-item pattern) + `eb9cff9` (facility switcher modal). The markup restructure is a real improvement, but **the three substantive bugs from my last review are still present** — I re-confirmed live on **M100054 (86 history rows)**:

**⚠️ 1. Row headline is still the constant `inventoryItemId` — not fixed.**
`<h2>{{ log.inventoryItemId }}</h2>` (line 102). All 86 rows are headlined **"M100062"** (verified: a single distinct `<h2>` value across the list). You added a date subtitle (good) and made the `:key` unique with `-${index}`, but the *prominent title* is still the same value on every row — and in my sample 7 of 8 rows also share the same timestamp ("Mar 22, 2026, 8:30 AM"), so the rows still read as identical. Headline something per-entry: `inventoryItemDetailSeqId`, `orderId`, or the resolved reason.

**⚠️ 2. Comment is still `log.description` → dead column — not fixed.**
`{{ log.description || "-" }}` (line 114). Confirmed **"-" on every row** — the inventory-detail records carry `reasonEnumId` (e.g. `VAR_TRFR_ORD_SHIPPED`), not `description`. Resolve `reasonEnumId` to its enum label (or show the order/shipment ref) instead.

**⚠️ 3. Loading state still uncovers the inventory fetch — not fixed.**
`onIonViewDidEnter` (lines 241–245) still does `isLoading = true → fetchProducts() → isLoading = false → fetchInventoryConfig() → fetchInventoryLogs()`, so QOH/ATP + history render their empty state during the config/logs fetch. Gate the cards/history on a flag that covers all three fetches.

**✅ New `FacilitySwitcherModal.vue` looks good (code review).**
Per-facility QOH/ATP joined from a batched parallel `productFacilities/search` (keyword + facilityId, mirroring the detail page's own fetch), searchbar, current-facility highlight, and a per-facility `try/catch` so one failed lookup degrades to `-` instead of breaking the list. I could **not exercise it live** — modal overlays are wedged in my current preview environment (app-wide, not this PR). 💡 It fires one request per facility on open (fine for ~15 facilities; consider batching if that grows), and `isLoading` isn't in a `finally` (only the inner per-facility catches save it — low risk).

**Verdict: 🔴 still needs another pass.** The restructure + switcher are nice, but the three data-binding fixes from the prior review are the blockers — and a "shared list-item pattern" that re-binds the same wrong fields doesn't resolve them.

_Re-audited the merge of `441-inventory-detail-restructure` with the integration base; bugs 1–2 re-confirmed live on M100054 (86 rows), bug 3 confirmed unchanged in code; new modal reviewed by code (live modal QA env-blocked)._

src/views/InventoryDetail.vue
dt2patel: Implemented + your virtual-product catch fixed, verified live on demo-oms (commit abd6eb0).

**Facility switcher** (your two comments): the inline `ion-select` is now a read-only `ion-item` showing the current facility + a **Change location** button (Settings end-slot pattern). It opens `FacilitySwitcherModal.vue` (modeled on `ManageFacilityGroupFacilitiesModal`): debounced `ion-searchbar` over name/id, `ion-list` of facilities with the current one marked, and each row shows this product's **live QOH/ATP** at that facility, fetched per facility in parallel on open (the search endpoint requires a facilityId).

**Virtual-product bug you flagged:** it was using `keyword: productId`, which fuzzy-matches a virtual product's *variants* and showed a variant's inventory as the product's. Both the modal and the detail's config fetch now use an **exact `productId` match**, so a virtual product truthfully shows no inventory and the modal/detail stay consistent. Verified: 10084 (virtual, V_antonia_racer_tank) → every facility "-" and detail QOH/ATP "-"; 10085 (variant) → real per-facility numbers (Broadway 1221/100, Central Warehouse −1227/99).

🤖 auto-addressed by the loop

dt2patel:
## QA + code review (shadow pass) — re-review (`abd6eb0`)

**`abd6eb0` (virtual-product inventory) — verified live, good fix. ✅**
The detail's `fetchInventoryConfig` (and the facility switcher) now query `productId=` (exact) instead of `keyword=` (fuzzy). Confirmed on demo-oms via the network: virtual product **10084** (`V_antonia-racer-tank`) → `productFacilities/search?productId=10084` (200) → QOH/ATP **"-"**, while stockable variant **10085** → `productId=10085` → **QOH 1221 / ATP 100**. So a virtual product no longer borrows a variant's stock, and the detail stays consistent with the switcher. Matches your commit's own verification.

**🔴 Still open — the 3 items from my previous re-review are unchanged** (this commit only touched the inventory-fetch param, not the history/loading):
1. History row headline is still `log.inventoryItemId` — constant per item (all 86 rows read "M100062" on M100054).
2. Comment still binds `log.description` → "-" on every row; the data carries `reasonEnumId`.
3. `onIonViewDidEnter` still sets `isLoading = false` before `fetchInventoryConfig()` / `fetchInventoryLogs()`, so the cards/history show an empty state during that fetch.

**Verdict: 🔴 still needs another pass** — the virtual-product fix is a genuine improvement, but the history identity/comment mapping + loading-state gating remain the blockers.

_Verified `abd6eb0` live (10084 virtual → "-", 10085 → 1221/100, both via exact `productId`, calls 200); bugs 1–3 re-confirmed unchanged in source._

PR #455, opened 2026-06-29
Implemented: facility counts + live net selected in pickup facility-group selector (#451)

### LINKED ISSUE #451: Show facility counts and live net selected facilities in pickup facility-group selector

dt2patel:
Resolves #451. Adds impact context to the facility-group selector (`AddProductFacilityGroupModal`) used by Store Pickup rules.

## Changes
- **Search bar** — filters groups by facility group name / id (local).
- **Selected groups as removable chips** below search, each with a close icon that removes it from the selection without closing the modal.
- **Per-group facility count** on each row (`ion-note slot="end"`), resolved via `fetchFacilitiesForGroup` and **cached** by `facilityGroupId` so filtering/scrolling never refetches.
- **Live net facilities total** for the current *unsaved* selection — dedupes by `facilityId` and subtracts excluded-group facilities from included-group facilities, matching `FacilityGroupImpactModal`. The side being edited uses the modal-local selection; the opposite side uses the already-applied groups.
- Opposite-side groups stay marked/unavailable (existing behaviour) and still show counts.

## Verified live on demo-oms (`hotwax.user`)
Rule M100153 → Include facility groups: search renders; per-group counts show (Online Facility Group 12, OMS Fulfillment 15, Ranked Warehouses 5, …); selecting "Online Facility Group" updates the net to "12 net facilities selected" and adds a removable chip; removing the chip returns it to "0". `npm run build` ✅.

Stacked on the integration branch `codex/integration-prs-428-434-437`. (Pairs with #450, the product-tag selector counterpart.)


src/components/AddProductFacilityGroupModal.vue
```diff
@@ -66,16 +88,68 @@ import EmptyState from '@/components/EmptyState.vue';
...
+  return facilityGroups.value.filter((group: any) =>
+    (group.facilityGroupName || '').toLowerCase().includes(q) || (group.facilityGroupId || '').toLowerCase().includes(q)
+  )
+})
+
+// Net = dedupe(facilities of included groups) minus facilities in any excluded group. The side being
+// edited uses the unsaved modal selection; the opposite side uses the already-applied groups.
+const netFacilityCount = computed(() => {
+  const includedGroups = props.type === 'included' ? selectedGroups.value : props.selectedFacilityGroups.included
+  const excludedGroups = props.type === 'excluded' ? selectedGroups.value : props.selectedFacilityGroups.excluded
```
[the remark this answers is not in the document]
dt2patel: Done in b51a15d — `netFacilityCount` now optional-chains the prop: `props.selectedFacilityGroups?.included || []` / `?.excluded || []`.

🤖 auto-addressed by the loop

src/components/AddProductFacilityGroupModal.vue
```diff
@@ -66,16 +88,68 @@ import EmptyState from '@/components/EmptyState.vue';
...
+  const excluded = facilityIdsFor(excludedGroups)
+  let net = 0
+  included.forEach((id) => { if (!excluded.has(id)) net++ })
+  return net
+})
+
+onMounted(async () => {
   selectedGroups.value = JSON.parse(JSON.stringify(props.selectedFacilityGroups[props.type]))
+  await loadFacilityCounts()
 })
```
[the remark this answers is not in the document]
dt2patel: Done in b51a15d — onMounted guards `props.selectedFacilityGroups?.[props.type]` before JSON.parse, defaulting to `[]`.

🤖 auto-addressed by the loop

src/components/AddProductFacilityGroupModal.vue
```diff
@@ -66,16 +88,68 @@ import EmptyState from '@/components/EmptyState.vue';
...
+  const groups = facilityGroups.value || []
+  if (!groups.length) return
+  isCounting.value = true
+  await Promise.all(groups.map(async (group: any) => {
+    if (facilitiesByGroupId.value[group.facilityGroupId]) return
+    const facilities = await productStore.fetchFacilitiesForGroup(group.facilityGroupId)
+    facilitiesByGroupId.value = { ...facilitiesByGroupId.value, [group.facilityGroupId]: facilities || [] }
+  }))
+  isCounting.value = false
+}
```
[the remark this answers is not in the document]
dt2patel: Done in b51a15d — `loadFacilityCounts` accumulates into a local map and assigns `facilitiesByGroupId` once instead of per group.

🤖 auto-addressed by the loop

dt2patel:

**Verdict: ✅ PASS** — clean, on-spec implementation of #451. Live modal UI QA was env-limited (note below); verdict is from a full code-review of the diff.

**Implements #451**
- **Search bar** + `filteredFacilityGroups` (matches group name or id). ✓
- Selected groups as **removable chips** (`removeSelectedGroup`). ✓
- **Per-row facility counts** (`facilityCountLabel` ← `facilitiesByGroupId`, populated by `loadFacilityCounts()`). ✓
- **Live net-facility count** for the unsaved selection: `netFacilityCount` dedupes the included groups' facilities into a Set and subtracts the excluded — using the modal-local selection for the side being edited and the applied groups for the other side. ✓ Correct net logic.
- Defensive guards (optional chaining on `selectedFacilityGroups?.[type]`), and `loadFacilityCounts` accumulates into a local map then assigns the ref once to avoid per-group reactivity churn (nicely commented). ✓
- `fetchFacilitiesForGroup` already exists in `atpProductStore` (base), so this PR **stands alone** — no dependency on #453. ✓

**💡 Optional polish**
- `loadFacilityCounts()` fires `fetchFacilitiesForGroup` for **every** group in parallel on mount (`Promise.all`). It caches (skips already-resolved), so it's once per open — but with many facility groups that's a burst of parallel calls. Consider lazy-loading counts for visible rows, or a single batched count query, if group counts grow large.

**🧪 Live QA note (env limitation, not a code fault)**
Opened from the Store Pickup rule editor's facility-group selector modal; like the sibling tag modal, modal data doesn't load in this cold preview (environment artifact). Verified by code review; re-verifiable live when the preview's modal data loads.

_Current with the integration base (0 behind); tip = merge result. All new i18n keys present (`Search facility groups`, `No result found for`, `{count} net facilities selected`)._

dt2patel:
## QA + code review (shadow pass) — re-audit follow-up

Re-audited this PR with a stricter, real-data method. **The original ✅ PASS holds** — the code-review claims check out under scrutiny (search/`filteredFacilityGroups`, removable chips, per-row counts via `facilitiesByGroupId`, and the `netFacilityCount` dedup-then-subtract math are all correct; it stands alone, no #453 dependency). Two items to add:

**⚠️ `loadFacilityCounts()` has no error handling — a single failed group fetch breaks the whole counts UI.**
```js
isCounting.value = true
await Promise.all(groups.map(async (group) => { … await productStore.fetchFacilitiesForGroup(…) … }))
facilitiesByGroupId.value = resolved   // ← never reached if any fetch rejects
isCounting.value = false               // ← never reached → spinner spins forever
```
If any one `fetchFacilitiesForGroup` rejects, `Promise.all` rejects and execution leaves the function before the two assignments: `isCounting` stays `true` (the `<ion-spinner>` spins indefinitely) and `facilitiesByGroupId` is never populated, so **every** row stays stuck on the `'…'` placeholder — and the rejection escapes `onMounted` unhandled. This isn't theoretical: during this session the demo backend was intermittently erroring (`runSolrQuery` 400s / transient network errors), which is exactly the trigger. Wrap the load in `try { … } finally { isCounting.value = false }` and catch per-group failures (e.g. `.catch(() => [])` inside the map) so one bad group degrades to a `0`/`—` count instead of taking down the whole modal.

**💡 `isAlreadyApplied()` is the one accessor that doesn't guard `selectedFacilityGroups`.**
`onMounted` and `netFacilityCount` use `?.`/`|| []`, but `isAlreadyApplied` does `props.selectedFacilityGroups[type].some(...)` unguarded. It's safe with the current sole caller (the editor always passes `{ included: [], excluded: [] }`), so this is latent — but worth matching the defensive style of the rest of the component.

**🧪 Live modal QA — still env-blocked, now confirmed NOT a fault of this PR.**
I could not get the facility-group selector to open in the preview to watch the counts/net render live. I verified this is **environmental, not a #455 regression**: on the same editor, the *tag-filter* modal — which this PR does not touch — also fails to open. So the live gap is a preview/session issue, not your modal. The counts + live net still want a clean-session live confirmation before merge, but nothing here points at the diff.

_Re-audited the merge of `451-pickup-fg-counts` with the integration base; full re-read of `AddProductFacilityGroupModal.vue`; modal-open failure isolated to the environment via the untouched tag modal._

src/components/AddProductFacilityGroupModal.vue
```diff
@@ -15,23 +15,42 @@
   </ion-header>
 
   <ion-content>
-    <ion-list v-if="facilityGroups.length">
-      <ion-item v-for="group in facilityGroups" :key="group.facilityGroupId"  @click="!isAlreadyApplied(group.facilityGroupId) ? updateSelectedGroups(group) : null">
-        <ion-label v-if="isAlreadyApplied(group.facilityGroupId)">{{ group.facilityGroupName }}</ion-label>
-        <ion-checkbox v-if="!isAlreadyApplied(group.facilityGroupId)" :checked="isAlreadySelected(group.facilityGroupId)">
-          {{ group.facilityGroupName ? group.facilityGroupName : group.facilityGroupId }}
+    <ion-searchbar :placeholder="translate('Search facility groups')" v-model="queryString" />
```
dt2patel: Same modal layout adjustment as the product-tag selector: move this searchbar into the modal header instead of rendering it as the first content row. Use a second `ion-toolbar` under the title/actions toolbar with the `ion-searchbar` inside it, leaving chips, net summary, and the list in `ion-content`.

src/components/AddProductFacilityGroupModal.vue
```diff
@@ -15,23 +15,42 @@
...
+
+    <div class="selected-chips" v-if="selectedGroups.length">
+      <ion-chip v-for="group in selectedGroups" outline :key="group.facilityGroupId">
+        <ion-label>{{ group.facilityGroupName || group.facilityGroupId }}</ion-label>
```
dt2patel: Please include the facility count in the selected chips as well, e.g. `Mall Locations: 12`. The rows expose the count, but once a group is selected the chip removes that impact context. Since `facilityCountLabel(group.facilityGroupId)` already exists, use it here and fall back gracefully while counts are loading.

src/components/AddProductFacilityGroupModal.vue
```diff
@@ -15,23 +15,42 @@
...
+    <!-- Live net facilities for the current (unsaved) selection: included groups' facilities minus excluded. -->
+    <ion-item lines="none" class="net-summary">
+      <ion-spinner v-if="isCounting" name="dots" slot="start" />
+      <ion-label color="medium">{{ translate("{count} net facilities selected", { count: netFacilityCount }) }}</ion-label>
+    </ion-item>
+
+    <ion-list v-if="filteredFacilityGroups.length">
+      <ion-item v-for="group in filteredFacilityGroups" :key="group.facilityGroupId" :button="!isAlreadyApplied(group.facilityGroupId)" @click="!isAlreadyApplied(group.facilityGroupId) ? updateSelectedGroups(group) : null">
+        <ion-label v-if="isAlreadyApplied(group.facilityGroupId)">{{ group.facilityGroupName || group.facilityGroupId }}</ion-label>
+        <ion-checkbox v-else :checked="isAlreadySelected(group.facilityGroupId)">
```
dt2patel: Please mirror the row ordering feedback from the product-tag modal: show the group label first, then the count, and put the checkbox/control at the far end. The current visual sequence is `checkbox -> label -> count`; it should read `label -> count -> checkbox` so users scan the group and facility impact first, with the selection control consistently at the end.

src/components/AddProductFacilityGroupModal.vue
dt2patel: Done in b7d1943. Moved the `ion-searchbar` into a second `ion-toolbar` in the modal header (under the title / Clear all toolbar); the chips, net-facilities summary, and group list remain in `ion-content`.

🤖 auto-addressed by the loop

src/components/AddProductFacilityGroupModal.vue
dt2patel: Done in b7d1943. Selected chips now include the facility count via `facilityCountLabel(group.facilityGroupId)` (e.g. `Mall Locations: 12`), falling back to the plain group name while the per-group counts are still loading.

🤖 auto-addressed by the loop

src/components/AddProductFacilityGroupModal.vue
dt2patel: Done in b7d1943. Reordered each group row to **label → count → checkbox**: `ion-label` for the group, `ion-note slot="end"` for the count, and `ion-checkbox slot="end"` as the final end-slot control. Already-applied groups show the count + the excluded/included note in place of the checkbox.

🤖 auto-addressed by the loop

dt2patel:
## QA + code review (shadow pass) — re-review (`b7d1943`)

Re-reviewed the new "Addressed review: FG modal row order, chip counts, searchbar in header" commit.

**✅ UI changes are correct (verified in code, mirrors the tag modal):**
- Each group row reordered to **label → count → checkbox**: `<ion-label>` name, then `<ion-note slot="end">` count, then `<ion-checkbox slot="end">` as the trailing control (only when not already-applied; applied rows keep the count + excluded/included note and stay non-interactive via `:button="!isAlreadyApplied(...)"`). The checkbox is a one-way `:checked="isAlreadySelected(...)"` indicator with the row click driving `updateSelectedGroups`, so it stays in sync. ✓
- Selected **chips now show the count** (`Name: 12`) once loaded, falling back to the plain name while counts resolve. ✓
- **Searchbar moved** into a second `ion-toolbar` in the header; chips/net-summary/list stay in `ion-content`. Ionic-native. ✓

**🔴 Still open — my prior ⚠️ on `loadFacilityCounts()` is not addressed.**
It still has no `try/catch/finally` (lines ~137–149): if any single `fetchFacilitiesForGroup` rejects, `Promise.all` throws before `facilitiesByGroupId.value = resolved` and `isCounting.value = false`, so the spinner spins forever and every row stays on the `'…'` placeholder. Given the demo backend's intermittent errors this is a real failure mode. Wrap in `try { … } finally { isCounting.value = false }` and catch per-group (`.catch(() => [])`) so one bad group degrades to `0`/`—` instead of taking down the whole modal.

**💡 `isAlreadyApplied()` still unguarded** — `props.selectedFacilityGroups[type].some(...)` with no `?.`/`|| []`, unlike `netFacilityCount`/`onMounted`. Safe with the current caller; worth matching the defensive style.

**🧪 Live:** I still couldn't open the facility-group modal in the preview to eyeball the new layout. Confirmed **environmental, not this PR**: the sibling tag modal (untouched here) also fails to present in the same session, and the console shows **no render/JS error** — only the known base `runSolrQuery tagsFacet` 400s (the bug #453 fixes). Layout verified by code review; a clean-session live pass is still wanted before merge.

**Verdict:** UI polish is good and on-spec; the count-load error handling (⚠️) is the one functional item left.

_Re-audited the merge of `451-pickup-fg-counts` (tip `b7d1943`) with the integration base; new-commit diff + current source read; modal-open failure isolated to the env via the unchanged tag modal._

dt2patel:
## QA + code review (shadow pass) — re-review **verified LIVE** (`b7d1943`)

Following up my code-review re-review: I got the FG modal rendered with real data (same env quirk as #453 — `modalController.create()` works but the modal's `present()` animation is wedged in this preview; I forced the created modal visible to inspect it). The new layout checks out on rule M100153:

- **Searchbar in the header toolbar** ("Search facility groups"). ✓
- **Row order label → count → checkbox** on every group (DOM-verified `ion-label` → `ion-note` → `ion-checkbox`). ✓
- **Per-group facility counts load and render**: Amazon Facility Group **1**, Online Facility Group **12**, OMS Fulfillment **15**, Pickup **12**, Ranked Warehouses **5**, Retail Stores **15**, Same Day Shipping **3**, Mall Locations **14**, … (12 groups total). ✓
- **Net summary renders**: "0 net facilities selected" (nothing selected). ✓

**⚠️ Re-confirming the one open item — `loadFacilityCounts()` failure handling.** Watching it live, the 12 per-group `fetchFacilitiesForGroup` calls take a few seconds to fill in — rows show the `…` placeholder with the spinner up until they resolve (here ~8s). On the happy path it resolves correctly. But because the load is a bare `await Promise.all(...)` with **no `try/catch/finally`**, a single rejected/hung group fetch would leave `isCounting` stuck `true` (spinner forever) and every row pinned on `…`. Wrap in `try { … } finally { isCounting.value = false }` and `.catch(() => [])` per group so one bad group degrades to `0`/`—`. (Also the 12-way fetch burst on open is a bit heavy — consider a single batched count query if the group list grows.)

**Verdict:** UI polish is correct and confirmed live; the count-load error handling (⚠️) is the remaining item before merge.

_Verified on a cold `:8212` preview of `451-pickup-fg-counts` merged with the integration base; modal content + counts inspected directly (DOM + screenshot) after the create-but-don't-animate env quirk._

dt2patel:
Addressed the open ⚠️ item (`loadFacilityCounts` failure handling) in def93e1.

`loadFacilityCounts()` now wraps the per-group load in `try { … } finally { … }` so `isCounting` is always cleared (no more permanent spinner / stuck `…` rows if a fetch hangs or rejects), and each `fetchFacilitiesForGroup` gets `.catch(() => [])` so one bad group degrades to an empty list (count `0`) instead of failing the whole `Promise.all`. Whatever resolved is still published in the `finally`.

Thanks for the live re-review confirming the layout (searchbar in header, label → count → checkbox, per-group counts, net summary). On the batched-count suggestion: agreed it's worth a single query if the group list grows — I'll leave that as a follow-up since today's per-group fetch matches what the rest of the modal flow uses.

🤖 auto-addressed by the loop

Issue #469, opened 2026-06-30
Inventory Detail: ATP/QOH balance cells show fabricated "0 → diff" totals (lastAvailableToPromise/lastQuantityOnHand not in API)


dt2patel:
## Summary

The ATP and QOH **balance-flow** cells on the Inventory Detail history accordion render a fabricated, corrupted "from → to" balance (e.g. `0 → -100` for a rollover whose ATP actually ended at `0`). The displayed "new" total is invented from a `0` baseline, not the real resulting balance.

## Where

`src/views/InventoryDetail.vue:208-220` (movement balance grid):

```vue
<span class="balance-flow">
  <span>{{ m.raw.lastAvailableToPromise ?? 0 }}</span>
  <ion-icon :icon="arrowForwardOutline" />
  <strong>{{ runningTotal(m.raw.lastAvailableToPromise, m.raw.availableToPromiseDiff) }}</strong>
</span>
```

`runningTotal` (`:502`):
```ts
function runningTotal(last: any, diff: any) {
  return (Number(last) || 0) + (Number(diff) || 0)
}
```

## Root cause

The cells reference `m.raw.lastAvailableToPromise` / `m.raw.lastQuantityOnHand`, **but the `oms/inventoryItem/detail` API never returns those keys.** Verified against live data (product `11084` @ `BROADWAY`): each row's keys are

```
inventoryItemId, inventoryItemTypeId, productId, datetimeReceived, facilityId,
locationSeqId, comments, quantityOnHandTotal, availableToPromiseTotal,
accountingQuantityTotal, inventoryItemDetailSeqId, effectiveDate,
quantityOnHandDiff, availableToPromiseDiff, accountingQuantityDiff,
reasonEnumId, description
```

`'lastAvailableToPromise' in row === false` for every row. So:

- Left side `{{ m.raw.lastAvailableToPromise ?? 0 }}` always renders `0`.
- Right side `runningTotal(undefined, diff)` computes `0 + diff`, i.e. it **recomputes a "new total" from a fake 0 baseline.**

For the `INV_ROLLOVER` row `O57403` (`availableToPromiseDiff: -100`, real `availableToPromiseTotal: 0`):
- True flow: `100 → 0`
- Rendered: `0 → -100` ← both numbers wrong; `-100` is neither the prior nor the resulting balance.

This affects **every** movement row's ATP and QOH cells, not just rollovers — they all show `0 → <diff>`.

## Expected behaviour

[...]

PR #470, opened 2026-07-01
Fix inventory history ATP/QOH delta display

dt2patel:
Closes #469

## Business summary
Inventory Detail history now reports ATP and QOH movement amounts truthfully. The expanded history cells no longer fabricate a `0 -> diff` balance flow from API fields that are not returned, which prevents operators from seeing impossible balances on rollover and adjustment rows.

## Changes
- Render ATP and QOH expanded history cells as signed movement deltas.
- Remove the unused `runningTotal` helper and all references to missing `lastAvailableToPromise` / `lastQuantityOnHand` fields.
- Keep Accounting behavior unchanged.

## Verification
- `pnpm --filter ./apps/order-routing build`
- Localhost browser check on `http://localhost:8104/inventory/11822`: expanded rollover row showed ATP/QOH signed deltas directly instead of a fabricated from/to flow.

## Base branch
This PR is stacked on `inventory-detail-page` as requested.


dt2patel:

**Verdict: ✅ PASS** — surgical, correct fix for #469. Reviewed the diff + the merged file on `codex/issue-469-inventory-history-deltas`.

**Why it's right**
- The expanded ATP/QOH history cells previously rendered `{{ lastAvailableToPromise ?? 0 }} → runningTotal(last, diff)`. Those `last*` fields are **not returned** by the inventory-detail API — they were read straight off `m.raw`, so the cells always displayed a fabricated **`0 → diff`** balance flow. This PR replaces them with the honest **signed movement delta** (`signed(...)` + `diffClass(...)`), which is the truthful representation. ✓
- **Accounting is correctly left unchanged** — that row legitimately has `accountingQuantityTotal` (a real returned running total), so keeping its `diff → total` display is right. Good that the fix scoped ATP/QOH only.
- This resolves exactly the *"other dimension renders `0 → <unchanged>`"* fabricated-flow issue I flagged on the Inventory Detail restructure (#447).

**Hygiene (Y8) — clean** ✓
- `runningTotal` helper removed with its last caller; **no dangling references remain** (grepped the merged `InventoryDetail.vue` — zero `runningTotal`/`lastAvailableToPromise`/`lastQuantityOnHand`), and `signed()`/`diffClass()` are defined and cover the new bindings. Tight `+2/-10`, single file, well-justified (Y2 ✓).

**⚠️ Y3 — attach before/after screenshots.** The body has a localhost verification note (`/inventory/11822`) but no images. A before/after of an expanded rollover row (old `0 → N` vs new `+N`/`-N`) is what Yash will ask for.

**🧪 Live re-verify (mine) was env-blocked, not a PR issue.** The shadow `:8212` preview is currently auth-dropped + blank-rendering (dev-server `getcwd` failure) after a long session — I couldn't drive it to re-confirm on a data-heavy product (e.g. M100054, 86 movement rows). Verdict rests on the diff + the field-shape corroboration above + your own 11822 check; a single expanded-row screenshot closes it.

_Reviewed `codex/issue-469-inventory-history-deltas` (tip e94cb3d) against base `inventory-detail-page`; merged-file grep confirms clean helper removal._

dt2patel:
## QA + code review (shadow pass) — ⚠️ CORRECTION, retracting my earlier PASS

My previous PASS on this PR was wrong. This PR **removes a legitimate feature** rather than fixing the actual bug, and I should have caught that — apologies.

**What's really going on**
The data model *does* support the expanded-row ATP/QOH running balance: `lastAvailableToPromise` / `lastQuantityOnHand` are real fields on the `oms/inventoryItem/detail` row and are populated on normal records — the base branch renders a correct `last → last+diff` flow for them (the `?? 0` fallback is itself the author signaling the field is expected). The defect is **only on legacy records** where those two fields are `null`: `{{ lastAvailableToPromise ?? 0 }}` renders `0` and `runningTotal(null, diff)` returns `0 + diff = diff`, so the cell fabricates a misleading **`0 → diff`** balance. So the fix should be a **null guard**, not deleting the balance flow for everyone.

**Why this PR is mis-scoped (🔴 request changes)**
It drops the `last → total` flow (and `runningTotal`) for **all** rows and shows only the signed delta — which is already shown in the collapsed summary row (`signed(availableToPromiseDiff)` at ~line 195). So on records that *do* have the data, it deletes the one thing the expanded view added (the actual before/after balance) and leaves the expanded cell redundant with the summary.

**Correct scope**
Guard each dimension independently and keep the feature where the data exists:
```vue
<!-- ATP -->
<span class="balance-flow" v-if="m.raw.lastAvailableToPromise != null">
  <span>{{ m.raw.lastAvailableToPromise }}</span>
  <ion-icon :icon="arrowForwardOutline" />
  <strong>{{ runningTotal(m.raw.lastAvailableToPromise, m.raw.availableToPromiseDiff) }}</strong>
</span>
<span v-else :class="diffClass(m.raw.availableToPromiseDiff)">{{ signed(m.raw.availableToPromiseDiff) }}</span>
```
…and the same, guarded separately, for `lastQuantityOnHand` / `quantityOnHandDiff`. Keep `runningTotal` (it's correct when the fields are present). Accounting stays as-is (it already has a real `accountingQuantityTotal`). Net effect: legacy rows no longer show the fake `0 → x`, and every non-legacy row keeps its real running balance.

**On my miss:** I over-trusted the PR's "these fields aren't returned" framing and fell back to code-review when the preview was down — and I ignored my own prior finding (#447) that these running balances rendered correctly against real data for a stocked product. Removal-vs-null-guard is exactly the scope call a review should catch. This needs another pass.

_Re-reviewed `codex/issue-469-inventory-history-deltas`; grounded in the base-branch balance-flow code + `oms/inventoryItem/detail` row shape (`useProductFacility.fetchInventoryLogs`)._

dt2patel:
Follow-up pushed for the corrected scope.

What changed:
- ATP/QOH now keep the `last -> runningTotal(last, diff)` balance flow when the API provides `lastAvailableToPromise` / `lastQuantityOnHand`.
- Legacy/null rows now fall back to the signed delta, so they no longer fabricate `0 -> diff`.
- Accounting remains unchanged.

Verification:
- `/Users/adityapatel/Documents/GitHub/accxui/node_modules/.bin/vite build` from isolated `order-routing-pr470-followup` worktree.

dt2patel:
## QA + code review (shadow pass) — re-review ✅ addressed

The re-push (`Guard inventory history balance flows`) implements exactly the null-guard scope — thanks for the quick turnaround.

- **Per-dimension guards**: `v-if="m.raw.lastAvailableToPromise != null"` (and independently for `lastQuantityOnHand`) — so ATP and QOH each keep the real `last → last+diff` running balance where the data exists, and fall back to the signed delta (`signed()` + `diffClass()`) only on legacy rows where the field is null. ✓
- **Fabrication removed**: dropped the `?? 0`, so legacy rows no longer render the misleading `0 → diff`. ✓
- `runningTotal` retained; Accounting cell unchanged. ✓ Verified helpers still defined in the merged head.

**Verdict: ✅ PASS.** The feature is preserved for populated records and the fabricated flow is gone for legacy ones. (Y3: a before/after screenshot of a legacy row (delta) vs a normal row (balance flow) is still nice to attach, but not blocking.)

_Re-reviewed `codex/issue-469-inventory-history-deltas` tip; diff + merged-file helper check._

ymaheshwari1:
Closing as all the commits in this branch are already on main.

Issue #473, opened 2026-07-01
Inventory search briefly shows 'No products found' while data is still loading


dt2patel:
## Current behavior
On the Find Inventory page (`/inventory`), the empty-state message renders immediately while the initial product-facility fetch is still in flight.

Users briefly see `No products found` before the real results arrive. This is misleading because the page is not actually empty yet; it is still loading.

## Why this is a problem
The current UI treats `products.length === 0` as a definitive empty state even during fetch.

That creates a false-negative flash on:
- initial page load
- facility changes
- search/filter changes
- pagination changes

## Expected behavior
While inventory rows are being fetched, the page should show a loading state instead of an empty state.

A better UX here is:
- show row-level skeleton placeholders while loading and there are no current results yet
- only show `No products found` after loading completes and the result set is truly empty

## Suggested implementation
In `src/views/Inventory.vue`:
1. Gate the empty state behind `!isLoading && !products.length`.
2. Add a loading-state branch for `isLoading && !products.length`.
3. Render a small list of skeleton rows that roughly matches the inventory-result layout so the page feels stable while data loads.

## Acceptance criteria
- `No products found` does not appear while the inventory request is still pending.
- The page shows a visible loading state or skeleton rows during the fetch.
- The empty state appears only after the request finishes and returns zero rows.
- Existing list rendering, select mode, filters, and pagination behavior remain unchanged.

PR #474, opened 2026-07-01
Add loading state to Inventory search (#473)

dt2patel:
## Business summary
The Inventory search page currently flashes a false empty state while records are still loading. This makes the page feel unreliable because users briefly see `No products found` before the real results arrive.

This PR adds a proper loading state so the inventory search experience reads as loading first, empty only when the fetch actually completes with no rows.

## What changed
- Adds skeleton inventory rows while the page is loading and there are no current results yet.
- Delays the `No products found` empty state until loading finishes.
- Keeps the existing list, filters, pagination, and select-mode behavior unchanged.

## Related issue
- Closes #473


dt2patel:
Closing this duplicate PR in favor of #475.

Reason: this branch includes a broad unrelated integration diff across many files and GitHub does not recognize it as closing #473, while #475 is a clean one-file implementation from `origin/main` that links `Closes #473` and has a passing build. I am not deleting the branch.

PR #475, opened 2026-07-01
Prevent inventory empty-state flash while loading

### LINKED ISSUE #473: Inventory search briefly shows 'No products found' while data is still loading

dt2patel:
Closes #473

Business summary:
The Find Inventory page now keeps users in a loading context while product-facility data is still being fetched, so users do not briefly see a false `No products found` result during initial load, facility changes, search changes, or pagination.

What changed:
- Added row-level skeleton placeholders for the empty-list loading state.
- Gated the empty state behind `!isLoading && !products.length`.
- Left the existing product list, selection, filters, and pagination behavior unchanged.

Verification:
- `/Users/adityapatel/Documents/GitHub/accxui/node_modules/.bin/vite build` from `/Users/adityapatel/Documents/GitHub/accxui/apps/order-routing-issue-473-clean`.

Backend blockers / limitations:
- None identified; this is a frontend rendering-state fix.

### DIFF
--- a/src/views/Inventory.vue
+++ b/src/views/Inventory.vue
@@ -24,7 +24,26 @@
             <ion-icon :icon="caretForwardOutline" />
           </ion-button>
         </div>
-        <p v-if="!products?.length" class="empty-state" data-testid="closed-empty-state">
+        <template v-if="showLoadingState">
+          <div class="list-item inventory-skeleton-row" v-for="row in loadingRows" :key="`inventory-skeleton-${row}`">
+            <ion-item>
+              <ion-thumbnail class="inventory-skeleton-thumbnail">
+                <ion-skeleton-text animated />
+              </ion-thumbnail>
+              <ion-label class="inventory-skeleton-copy">
+                <ion-skeleton-text animated class="inventory-skeleton-line inventory-skeleton-line-primary" />
+                <ion-skeleton-text animated class="inventory-skeleton-line inventory-skeleton-line-secondary" />
+              </ion-label>
+            </ion-item>
+            <div v-for="column in 5" :key="`inventory-skeleton-${row}-${column}`">
+              <ion-label>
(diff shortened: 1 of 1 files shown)


dt2patel:
Addressed the skeleton layout review feedback.

What changed:
- Added a disabled checkbox placeholder to skeleton rows so the thumbnail/label columns line up with the loaded inventory rows and avoid horizontal shift.

Verification:
- `/Users/adityapatel/Documents/GitHub/accxui/node_modules/.bin/vite build` from `/Users/adityapatel/Documents/GitHub/accxui/apps/order-routing-issue-473-clean`.

PR #499, opened 2026-07-17
Prepare consolidated order routing for production

dt2patel:
## Summary

- Replaces closed PR #495 with a current `main`-based production-readiness proposal for the consolidated order-routing list and detail experience.
- Makes the canonical detail page support regular routing navigation/editing, isolated simulation variations, and Circuit proposals through the same routing JSON and shared card components.
- Adds explicit working-copy, dirty-state, card-level accept/reject, save/discard, unsaved-navigation, loading, modal-data, and error/empty-state contracts.
- Lets Circuit inspect and propose edits against the active simulation variation without mutating the live baseline.
- Adds a baseline comparison workflow that groups added, removed, changed, and reordered settings by routing section, with section-level and full-variation reset controls.
- Completes the variation lifecycle with baseline simulation runs, variation create/update/discard, interrupted-run recovery, stale-operation guards, partial-failure results, and persisted result navigation.
- Isolates cached routing state by OMS instance and user session, clears routing/Circuit/simulation state on context changes, and safely redirects stale direct-detail routes.
- Stabilizes direct detail entry with a structural card skeleton, canonical backend filter keys, human-readable status fallbacks, clearer reorder controls, and visible chat/resize feedback.
- Fails Circuit, simulation, and Test Drive integrations closed unless their deployment trust settings are explicit; removes production mock fallback behavior.
- Enables the AccxUI-owned IDE Trace Vite plugin for source-aware local browser review without duplicating the workspace dependency.

Closes #498.

## Business impact

Routing administrators get one dependable workflow for inspecting and editing normal routing, testing alternative configurations without touching live routing, reviewing Circuit changes in their exact UI context, and returning to saved simulation results. A variation now explains exactly how it differs from the live baseline and lets an administrator undo one section or the full variation before running it. Initial detail loads hold the final card geometry, routing statuses remain readable when reference data is late, filter edits update the canonical saved condition, and Circuit's empty, loading, proposal-review, and resizing states are clear.

## Root causes addressed

- Direct detail navigation initially rendered an empty/spinner canvas before the editor columns existed, producing a dramatic layout shift.
- Filter controls use enum names while persistence uses backend field codes; writing the UI name could create a duplicate condition that the control would not read back.
- Missing or delayed reference data could leak status system IDs into the interface.
- The chat panel had no empty-state guidance and its resize separator was difficult to distinguish from the canvas boundary.
- Routing detail state was persisted without an OMS/user session boundary, allowing stale group data to appear after a login or backend-context change.
[...]

PR #511, opened 2026-07-21
⚡ perf: Optimize fetchRoutingsInformation concurrency

dt2patel:
💡 **What:** Replaced the `forEach(async ...)` loop with `Promise.all(group.value.routings.map(async ...))` in `fetchRoutingsInformation`.
🎯 **Why:** The previous code fired off asynchronous API requests for each routing item but did not await their completion before the `fetchRoutingsInformation` function returned. This caused a race condition where the function appeared to finish but the network requests were still running in the background. Using `Promise.all()` with `map()` ensures all asynchronous calls run concurrently and the function correctly waits for all of them to resolve before proceeding.
📊 **Measured Improvement:** Unable to provide numerical benchmark results due to a missing monorepo workspace configuration (`catalog:` specifiers and `eslint`/`vitest` configuration errors prevented local script execution). However, based on the codebase structure, this clearly fixes a latent race condition (unawaited promises) and successfully implements the `TODO: update logic to fetch the routings in parallel` requirement in a standard, optimal manner by enabling concurrent API dispatch instead of serial/unawaited processing.

---
*PR created automatically by Jules for task [9285621602173744336](https://jules.google.com/task/9285621602173744336) started by @dt2patel*

### DIFF
--- a/src/views/BrokeringRunTest.vue
+++ b/src/views/BrokeringRunTest.vue
@@ -200,8 +200,7 @@ async function fetchRoutingGroupInformation() {
 }
 
 async function fetchRoutingsInformation() {
-  // TODO: update logic to fetch the routings in parallel
-  await group.value.routings.forEach(async (routing: any) => {
+  await Promise.all(group.value.routings.map(async (routing: any) => {
     let route = {} as any
     try {
       const resp = await api({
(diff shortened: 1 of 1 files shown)


dt2patel:
Closing as **superseded by #499** (consolidation refactor). #499 already parallelizes this logic: `fetchRoutingsInformation` in the renamed `src/views/RoutingGroupTest.vue` (line ~279) already uses `Promise.all(routings.map(async ...))`. Porting this fix onto the #499 line would be a no-op. If #499 is abandoned, reopen — the fix is still valid on `main`.

PR #519, opened 2026-07-25
⚡ Bolt: Parallelize enum fetching in utilStore

dt2patel:
💡 **What:**
Replaced a sequential `for...of` loop in `fetchRoutingEditorEnums` with `Promise.all` to fetch configuration enums concurrently. Also modified `fetchEnums` to safely merge updates directly into the store's global state (`this.enums`) at the type level inside the response handler block, rather than maintaining a function-scope snapshot that caused race conditions.

🎯 **Why:** 
Previously, the editor configuration enums were fetched one by one because the underlying `fetchEnums` action was susceptible to race conditions (it snapshotted the state before making the network request, then merged and saved it, overwriting other in-flight updates). This sequential fetching created an unnecessary network waterfall that delayed the initialization of the routing rules editor.

📊 **Impact:** 
Eliminates a sequential network waterfall consisting of 5 consecutive API calls, significantly reducing load time for the editor views by allowing these queries to run concurrently.

🔬 **Measurement:** 
1. Open the network tab in browser devtools.
2. Navigate to a route that loads the order routing editor (which triggers `fetchRoutingEditorEnums`).
3. Observe that the `/admin/enums` calls for the different enum type IDs are now dispatched in parallel rather than sequentially, reducing total request latency.

---
*PR created automatically by Jules for task [10313912425221578762](https://jules.google.com/task/10313912425221578762) started by @dt2patel*


ymaheshwari1:
Closing as duplicate of #521

PR #518, opened 2026-07-24
⚡ Bolt: [performance improvement] Parallelize fetching in fetchRoutingEditorEnums

dt2patel:
💡 **What:** Optimized `fetchRoutingEditorEnums` in `utilStore.ts` to fetch multiple enum types concurrently using `Promise.all`. To support this safely, I refactored `fetchEnums` and `fetchOmsEnums` to merge results directly into `this.enums` after the network request resolves, avoiding race conditions that previously necessitated sequential fetching.

🎯 **Why:** The sequential `for...of` loop was creating an N+1 query bottleneck, forcing the application to wait for each enum family to finish loading before starting the next. This unnecessarily increased latency and blocked initialization.

📊 **Impact:** Significantly reduces the load time for initializing editor resources by downloading all required enum families in parallel.

🔬 **Measurement:** Verify the optimization by checking the network tab in browser developer tools when the application loads the enums. You should observe the requests for `admin/enums` happening simultaneously rather than in a stepped waterfall.

---
*PR created automatically by Jules for task [12098671546531904890](https://jules.google.com/task/12098671546531904890) started by @dt2patel*

PR #520, opened 2026-07-26
⚡ Bolt: parallelize enum fetching in util store

dt2patel:
💡 **What:**
Replaced the sequential `for...of` loop in `fetchRoutingEditorEnums` with a concurrent `Promise.all` mapping. Additionally, refactored state assignment in `fetchEnums` and `fetchOmsEnums` to merge data directly into the live `this.enums` state *after* the network request completes, rather than mutating a snapshot captured before the network request.

🎯 **Why:**
The application was making 5 sequential network requests to load required editor enums on startup, creating an unnecessary network waterfall that blocked execution. Attempting to parallelize these requests naively caused a race condition where the last-resolving request would overwrite the state of the earlier-resolving requests due to stale closures. This patch safely parallelizes the requests and fixes the underlying race condition.

📊 **Impact:**
Reduces the enum loading time from the sum of 5 sequential API requests down to the latency of a single (the slowest) API request. This significantly accelerates the time-to-interactive for the routing editor.

🔬 **Measurement:**
Open the network tab when navigating to the routing editor. You should see all 5 requests to `admin/enums` trigger simultaneously rather than cascading sequentially. Verify that the routing editor still correctly renders all filter/sort dropdowns using the loaded enums.

---
*PR created automatically by Jules for task [6419755223225285071](https://jules.google.com/task/6419755223225285071) started by @dt2patel*


ymaheshwari1:
Closing as duplicate #518

PR #555, opened 2026-08-14
feat: include systemMessageRemoteId in rule group update payloads for backward support

adarshGit-hub:


ymaheshwari1:
Closing as we are assuming when atp is used via company, the instance is on latest release and thus this backward co patibility changes are not required.

